PR #4, opened 2023-02-16
Telemetry Manager code move to new repository

hisarbalik:
**Description**

- Move entire Telemetry component code from Kyma repository to the new Telemetry Manager repository

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/16745


Makefile
```diff
@@ -1,9 +1,11 @@
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= telemetry-operator:latest
```
chrkl, line 3: let's rename this also to telemetry-manager

Makefile
```diff
@@ -96,14 +108,6 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
-.PHONY: deploy
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
chrkl, line 100: I'm wondering if the deploy and underlay targets might become useful again for modularisation. When getting rid of Helm, this might be the main flow for in-cluster testing.

chrkl (commented):
The linked issue the PR description seems to be the wrong one.

Makefile
hisarbalik: image name renamed to telemetry manager

Makefile
hisarbalik: sure, this can be default flow for deploy and undeploy after creating a module from controller

chrkl: So should we keep it or remove first and add again when needed?

hisarbalik: they will not work any way for now, lets remove and add the both target for module

chrkl (approved):

PR #19, opened 2023-02-21
Adding pod ip usage for trace collector

Teneroy:
**Description**

- Added protocols struct
- Added new environment variable to telemetry-trace-collector indicating pod ip
- Injected pod ip into telemetry-trace-collector in order to fix localhost related warnings

**Related issue(s)**

 https://github.com/kyma-project/kyma/issues/16736


controller/tracepipeline/config_builder.go
```diff
@@ -41,9 +41,13 @@ type ExporterConfig struct {
 	Logging  LoggingExporterConfig `yaml:"logging,omitempty"`
 }
 
+type OTLP struct {
```
chrkl, line 44: Let's call it `OTLPReceiverConfig` since we already have an `OTLPExporterConfig`

controller/tracepipeline/config_builder.go
```diff
@@ -41,9 +41,13 @@ type ExporterConfig struct {
...
+	Protocols map[string]any `yaml:"protocols,omitempty"`
```
chrkl, line 45: Make this (and the nested properties) also a strict type.

chrkl (commented):
Also create a PR to bump the image in the telemetry chart to run the integration tests.

chrkl (approved):

PR #43, opened 2023-03-01
Remove sleep statement from webhook test

chrkl:
**Description**

Sleep statement increases the test runtime significantly and does not have an obvious use.

- Remove sleep statement from webhook test

**Related issue(s)**


skhalash (approved):

chrkl:
@skhalash would like to run the unit tests on Prow a few times and ensure that it is not introducing any flakiness before merging.

PR #46, opened 2023-03-02
Remove unnecessary cli flags

rakesh-garimella:
**Description**

- Remove unnecessary cli flags
- Add the default values as const

**Related issue(s)**


main.go
```diff
@@ -361,45 +323,25 @@ func main() {
...
-	if fluentBitConfigMap == "" {
-		return errors.New("--fluent-bit-cm-name flag is required")
-	}
-	if fluentBitSectionsConfigMap == "" {
-		return errors.New("--fluent-bit-sections-cm-name flag is required")
-	}
-	if fluentBitParsersConfigMap == "" {
-		return errors.New("--fluent-bit-parser-cm-name flag is required")
-	}
 	if fluentBitDaemonSet == "" {
```
dennis-ge, line 326: this can be removed as it is always false

main.go
```diff
@@ -361,45 +323,25 @@ func main() {
...
 		return errors.New("--fluent-bit-ds-name flag is required")
 	}
-	if fluentBitEnvSecret == "" {
-		return errors.New("--fluent-bit-env-secret flag is required")
-	}
-	if fluentBitFilesConfigMap == "" {
-		return errors.New("--fluent-bit-files-cm flag is required")
-	}
 	if telemetryNamespace == "" {
```
dennis-ge, line 329: this can be removed as it is always false

dennis-ge (commented):
Nice PR :) some small comments

config/manager/manager.yaml
```diff
@@ -44,7 +44,6 @@ spec:
       - command:
         - /manager
         args:
-        - --leader-elect
         - --cert-dir=/tmp
         - --fluent-bit-cpu-limit=1
         - --fluent-bit-memory-limit=1Gi
```
dennis-ge, line 49: `fluent-bit-cpu-limit`,  `fluent-bit-memory-limit`, `fluent-bit-denied-output-plugins`, `fluent-bit-max-pipelines` and all trace-collector flags could also be removed as we use the default value.

Issue #36, opened 2023-02-27
Implement module contract in telemetry-manager


a-thaler:
**Description**
We need to implement the module interface

**Criterias**
- Manager deployed by helm must have the feature disabled, also no new CRD gets deployed
- Manager deployed via kustomize or manual will have the feature enabled and CRD deployed
- CRD must be conform to the requirements of the modularization concept and will have no spec
- Manager generates the default CR if not present (ignore behaviour if further CRs exists for now)
- CR status (predefined by module concept) should reflect the manager readiness, for now let's reflect the status of the webhook endpoint
- Have first tests assuring that a status gets produced

References:
- Reference implementation -> https://github.com/kyma-project/template-operator
- Example with manifest library -> https://github.com/kyma-project/module-manager/#manifest-library
- Example implementations -> https://github.com/kyma-project/nats-manager and https://github.com/kyma-project/keda-manager

PR #67, opened 2023-03-09
Add module contract in telemetry manager

hisarbalik:
**Description**

- Initial module contract added

**Related issue(s)**
https://github.com/kyma-project/telemetry-manager/issues/36

### DIFF
--- a/controllers/operator/telemetry_controller_test.go
+++ b/controllers/operator/telemetry_controller_test.go
@@ -46,7 +46,6 @@ var _ = Describe("Deploying a Telemetry", func() {
 
 				return nil
 			}, timeout, interval).Should(BeNil())
-
 			Expect(k8sClient.Delete(ctx, telemetryTestObj)).Should(Succeed())
 		})
 	})


controllers/telemetry/telemetrymanager_controller.go
```diff
@@ -0,0 +1,184 @@
...
+//+kubebuilder:rbac:groups=telemetry.kyma-project.io,resources=telemetrymanagers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=telemetry.kyma-project.io,resources=telemetrymanagers/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=telemetry.kyma-project.io,resources=telemetrymanagers/finalizers,verbs=update
+
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	objectInstance := telemetryv1alpha1.TelemetryManager{}
```
shorim, line 65: I think we should move the reconciliation logic into the `internal/reconciler` dir to be consistent with the changes done here #64

apis/telemetry/v1alpha1/telemetrymanager_types.go
```diff
@@ -0,0 +1,131 @@
...
+	condition.ObservedGeneration = objGeneration
+	meta.SetStatusCondition(&s.Conditions, *condition)
+	return s
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+// TelemetryManager is the Schema for the telemetrymanagers API
+type TelemetryManager struct {
```
shorim, line 100: don't forget to rename it to `Telemetry` as well as the other types

controllers/telemetry/telemetrymanager_controller_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Deploying a TelemetryManager", func() {
+	const (
+		timeout  = time.Second * 100
```
shorim, line 17: I think that this is a typo and you intended to have the timeout as 10 seconds instead of 100 seconds, right?

shorim: Okay, I have noticed now that in the controller test for LogPipeline and LogParser the timeout is 10s, but in the TracePipeline the timeout is 100s. Is there a reason for this big difference in the timeout?

controllers/telemetry/suite_test.go
```diff
@@ -130,6 +130,10 @@ var _ = BeforeSuite(func() {
 	err = tracepipelineReconciler.SetupWithManager(mgr)
 	Expect(err).ToNot(HaveOccurred())
 
+	telemetryManagerReconciler := NewTelemetryManagerReconciler(client, mgr.GetScheme(), mgr.GetEventRecorderFor("dummy"))
+
```
shorim, line 134: a little nitpick, just to keep the same spacing structure as for the other reconcilers, can you just remove this empty line and have an empty line between `err = telemetryManagerReconciler.SetupWithManager(mgr)` and  `go func() {...}()`?

apis/telemetry/v1alpha1/telemetrymanager_types.go
hisarbalik: renamed

shorim (approved):

PR #92, opened 2023-03-21
Update Makefile 'Build Dependency' targets

dennis-ge:
**Description**

- Recently the `Build Dependency` make targets failed sometimes because a binary is already present in the local `bin` directory.
- Since this is quite annoying, I checked whether our Makefile is outdated and found some updates to those make targets:  

1. There is now an if section checking if the version is outdated: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L142-L145. This is not the reason for the failing make targets but I anyway changed it because it is handy.

    ```shell
	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then \
		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
		rm -rf $(LOCALBIN)/kustomize; \
	fi
   ```

3.  There is now a check if the binary already exists before it is reinstalled: https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile#L146. This should fix the failing make targets
  ```shell
  test -s $(LOCALBIN)/kustomize || { curl -Ss $(KUSTOMIZE_INSTALL_SCRIPT) --output install_kustomize.sh && bash install_kustomize.sh $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN); rm install_kustomize.sh; }
  ```

**Related issue(s)**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,34 +137,44 @@ GINKGO ?= $(LOCALBIN)/ginkgo
 K3D ?= $(LOCALBIN)/k3d
 
 ## Tool Versions
-KUSTOMIZE_VERSION ?= v5.0.0
+KUSTOMIZE_VERSION ?= v5.0.1
 CONTROLLER_TOOLS_VERSION ?= v0.11.3
 K3D_VERSION ?= v5.4.7
-GINKGO_VERSION ?= v2.9.1
+GINKGO_VERSION ?= v2.9.0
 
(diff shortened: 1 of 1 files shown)


skhalash:
It was quite annoying indeed, thanks!

skhalash (approved):

PR #93, opened 2023-03-21
Restrict RBAC for telemetry manager

rakesh-garimella:
**Description**

- Create new cache which filters resources only from the given namespace. In order to restrict the RBAC for telemetry 
manager. 
- The cache contains resources that are filtered by namespace. Additionally secret resource is not namespace restricted. As we would like read secrets from different namespaces.
**Related issue(s)**

https://github.com/kyma-project/kyma/issues/17001

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,9 +19,13 @@ package main
 import (
 	"errors"
 	"flag"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"net/http"
 	"os"
 	"path"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -291,6 +290,17 @@ func main() {
 		LeaderElectionNamespace: telemetryNamespace,
 		LeaderElectionID:        "cdd7ef0b.kyma-project.io",
 		CertDir:                 certDir,
+		NewCache: cache.BuilderWithOptions(cache.Options{
+			SelectorsByObject: cache.SelectorsByObject{
+				&corev1.Secret{}:         {},
+				&appsv1.Deployment{}:     {Field: fields.SelectorFromSet(fields.Set{"metadata.namespace": telemetryNamespace})},
```
skhalash, line 296: Could you please extract `fields.SelectorFromSet(fields.Set{"metadata.namespace": telemetryNamespace})` into a helper function?

skhalash (approved):

PR #118, opened 2023-04-05
Refactor trace generation in e2e tests

skhalash:
**Description**

Tracing e2e tests generate the data and then verify that it's been delivered to the mock backend. Currently, the OTEL Tracing SDK is used to do it. It has, however, a major drawback: the SDK is complex on its own and has a lot of config parameters. It also looks completely different from the Metric SDK. This PR refactors the tests to use [`pdata` data structs](https://github.com/open-telemetry/opentelemetry-collector/tree/main/pdata) directly and sends them to the OTLP endpoint. 
It is the lowest lever you can get and gives direct control over the data getting sent. Another advantage is that it looks very symmetric for metrics, traces, and logs.

- Remove dependency to OTEL Tracing SDK
- Use ptrace data structures, as well as it's JSON marshalling capabilities
- Merge trace generating and verification into a single spec
- Write tests for the custom matchers

**Related issue(s)**


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: eu.gcr.io/kyma-project/telemetry-manager
-  newTag: v20230329-e1d53cad
+  newName: k3d-kyma-registry:5000/telemetry-manager
```
ameteiko: As you reverted this change in the previous commit, I assume this should be removed.

internal/otelmatchers/trace_matchers.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/onsi/gomega/types"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+func ConsistOfSpansWithIDs(expectedSpanIDs []pcommon.SpanID) types.GomegaMatcher {
+	return gomega.WithTransform(func(actual interface{}) ([]pcommon.SpanID, error) {
+		actualBytes, ok := actual.([]byte)
+		if !ok {
+			return nil, fmt.Errorf("ConsistOfSpansWithIDs rqquires a []byte, but got %T", actual)
```
ameteiko, line 18: Typo: rqquires -> requires

internal/otelmatchers/trace_matchers.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}
+
+	return spans
+}
+
+func unmarshalOTLPJSONTraces(buf []byte) ([]ptrace.Traces, error) {
+	var results []ptrace.Traces
+
+	tracesUnmarshaler := &ptrace.JSONUnmarshaler{}
```
ameteiko, line 101: I see no reason for using a reference here. Can be simplified down to:

```go
tracesUnmarshaler := ptrace.JSONUnmarshaler{}
```

internal/otelmatchers/trace_matchers.go
```diff
@@ -0,0 +1,116 @@
...
+					spans = append(spans, scopeSpans.Spans().At(k))
+				}
+			}
+		}
+	}
+
+	return spans
+}
+
+func unmarshalOTLPJSONTraces(buf []byte) ([]ptrace.Traces, error) {
```
ameteiko, line 98: I'd prefer to have a unit test for this function.

internal/otelmatchers/trace_matchers.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+		return actualSpanIDs, nil
+	}, gomega.ConsistOf(expectedSpanIDs))
+}
+
+func EachHaveTraceID(expectedTraceID pcommon.TraceID) types.GomegaMatcher {
+	return gomega.WithTransform(func(actual interface{}) ([]pcommon.TraceID, error) {
+		actualBytes, ok := actual.([]byte)
+		if !ok {
+			return nil, fmt.Errorf("EachHaveTraceID rqquires a []byte, but got %T", actual)
```
ameteiko, line 40: Typo: rqquires -> requires

internal/otelmatchers/trace_matchers.go
```diff
@@ -0,0 +1,116 @@
...
+
+		var actualSpanIDs []pcommon.SpanID
+		for _, span := range actualSpans {
+			actualSpanIDs = append(actualSpanIDs, span.SpanID())
+		}
+		return actualSpanIDs, nil
+	}, gomega.ConsistOf(expectedSpanIDs))
+}
+
+func EachHaveTraceID(expectedTraceID pcommon.TraceID) types.GomegaMatcher {
```
ameteiko, line 36: The name for the `EachHaveTraceID` function is too vague to me. As it is called for an HTTPBody, it is unclear what `Each...` implies. I suggest aligning it with the `ConsistOfSpansWithIDs` one.

internal/otelmatchers/trace_matchers.go
```diff
@@ -0,0 +1,116 @@
...
+	"bytes"
+	"fmt"
+
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+func ConsistOfSpansWithIDs(expectedSpanIDs []pcommon.SpanID) types.GomegaMatcher {
```
ameteiko, line 14: I wonder if we can use go generics for both `ConsistOfSpansWithIDs` and `ConsistOfSpansWithIDs`?

test/e2e/tracing_test.go
```diff
@@ -104,32 +91,30 @@ var _ = Describe("Tracing", func() {
...
+			attrs.PutStr("attrA", "chocolate")
+			attrs.PutStr("attrB", "raspberry")
+			attrs.PutStr("attrC", "vanilla")
+
+			traces := makeTraces(traceID, spanIDs, attrs)
 
-		It("Should verify traces were sent to the backend", func() {
+			sendTraces(context.Background(), traces, "localhost", 4317)
+
+			attrs.PutBool("bool", true)
```
ameteiko: Is attribute modification _after_ sending the traces intentional?

config/manager/kustomization.yaml
skhalash: What exactly do you mean?

internal/otelmatchers/trace_matchers.go
skhalash: Agree

test/e2e/tracing_test.go
skhalash: Nope, that change was reverted

config/manager/kustomization.yaml
skhalash: It was reverted in a follow-up commit

ameteiko: [This commit](https://github.com/kyma-project/telemetry-manager/pull/118/commits/c809beacd8d52f2c86519a5fa9924b78037a9dd2) reverted these exact changes.

skhalash: Yup, exactly. So what has to be removed?

internal/otelmatchers/trace_matchers.go
skhalash: I don't like the names, to be honest. Changed them, please check.

internal/otelmatchers/trace_matchers.go
skhalash: Please elaborate :)

config/manager/kustomization.yaml
ameteiko: My bad, it's all good for this change, please ignore my comments above.

internal/otelmatchers/trace_matchers.go
skhalash: I suggest maybe we test every matcher (`ConsistOf...`) the way Gomega suggests: https://onsi.github.io/gomega/#testing-custom-matchers. It's more of an integration test that would implicitly test the unmarshal function. What do you think?

internal/otelmatchers/trace_matchers.go
skhalash: I would not say generics would solve it, we could instead introduce another helper to reduce code duplication, but not sure it's worth it

internal/otelmatchers/trace_matchers_test.go
```diff
@@ -0,0 +1,240 @@
...
+		})
+	})
+
+	Context("with invalid input", func() {
+		BeforeEach(func() {
+			fileBytes = []byte{1, 2, 3}
+		})
+
+		It("should error", func() {
+			_, err := ConsistOfSpansWithIDs(expectedSpansWithIDs).Match(fileBytes)
```
ameteiko, line 85: To fully test the contract of the Function Under Test (`ConsistOfSpansWithIDs`), I suggest to ensure that the returned list of span IDs is nil is error has occured:

```go
spans, err := ConsistOfSpansWithIDs(expectedSpansWithIDs).Match(fileBytes)

Expect(err).Should(HaveOccurred())
Expect(spans).ToBeNil()
```

internal/otelmatchers/trace_matchers_test.go
```diff
@@ -0,0 +1,240 @@
...
+	BeforeEach(func() {
+		expectedSpansWithIDs = []pcommon.SpanID{[8]byte{1}, [8]byte{2}, [8]byte{3}}
+	})
+
+	Context("with nil input", func() {
+		BeforeEach(func() {
+			fileBytes = nil
+		})
+
+		It("should fail", func() {
```
ameteiko, line 31: I'd prefer simpler one:

```go
Context("with nil input", func() {
    It("should fail", func() {
        Expect(nil).ShouldNot(ConsistOfSpansWithIDs(expectedSpansWithIDs))
    })
})
```

skhalash: Agree, but it's not the same test. The function param is of type `interface{}`, so passing a nil directly is not the same as passing a nil slice. I added both tests

internal/otelmatchers/trace_matchers_test.go
skhalash: The `Match` method does not return spans, it always returns a `success` flag, which is then internally checked by Gomega (you would never call `Match` explicitly). I added a check for the flag.

ameteiko (approved):

PR #126, opened 2023-04-13
Enhance pipeline lock mechanism

chrkl:
**Description**

Enhance pipeline lock mechanism to support limiting the number of TracePipelines and MetricPipelines to a number that is larger than one. The enhanced lock implementation uses the owner list to identify the current lock holders.

- Support limiting the pipeline count to a number larger than 1

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/17223

### DIFF
--- a/internal/kubernetes/lock.go
+++ b/internal/kubernetes/lock.go
@@ -13,13 +13,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-func TryAcquireLock(ctx context.Context, client client.Client, lockName types.NamespacedName, owner metav1.Object) error {
+var errLogInUse = errors.New("lock is already acquired by other resources")
+
+type ResourceCountLock struct {
+	client    client.Client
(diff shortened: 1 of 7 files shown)


internal/kubernetes/lock.go
```diff
@@ -13,13 +13,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-func TryAcquireLock(ctx context.Context, client client.Client, lockName types.NamespacedName, owner metav1.Object) error {
```
dennis-ge, line 16: Is it intended that variable is named `errLogInUse`  and not `errLockInUse`? 

I also would rather just have the err msg as constant and then create the error in the method. But also fine declaring the err here.

dennis-ge: This comment is about this line 

```
var errLogInUse = errors.New("lock is already acquired by other resources")
```

dennis-ge (approved):

Issue #113, opened 2023-04-03
[Telemetry] Module verification and release process


a-thaler:
**Description**
The telemetry module is prepared for modularization and can be installed following this guide: https://github.com/kyma-project/telemetry-manager/blob/main/docs/deploying-module-template.md

Now we need to ensure that this integration is working at any time, so automate these steps on base of a k3d cluster and the local docker registry. The test should be executed on every PR.
Additionally, we need a way to release a module that executes the same test but additionally creates a GitHub release

**Criteria**
- Have a new prow job, automating the steps being documented on every PR
- Have a new prow job triggered on tag creation which will do:
  - perform the same tests as above
  - generates a module template
  - uploads the OCI images for that version
  - creates a GitHub release entry with the module template as yaml

PR #162, opened 2023-05-07
Restructure module targets in Makefile

shorim:
**Description**

- Remove `run-with-lm` sub-targets from Makefile and move their logic into a separate bash script and rename it to `e2e-deploy-module`
- Remove `release` sub-targets from Makefile and move their logic into a separate bash script
- Update README to use the new the new `e2e-deploy-module` target instead of pointing to documentation of manual steps

**Related issue(s)**

#113


Makefile
```diff
@@ -296,11 +207,11 @@ define os_error
 $(error Error: unsupported platform OS_TYPE:$1, OS_ARCH:$2; to mitigate this problem set variable KYMA with absolute path to kyma-cli binary compatible with your operating system and architecture)
 endef
 
-KYMA_FILE_NAME ?=  $(shell ./hack/get_kyma_file_name.sh ${OS_TYPE} ${OS_ARCH})
+KYMA_FILENAME ?=  $(shell ./hack/get-kyma-filename.sh ${OS_TYPE} ${OS_ARCH})
 KYMA_STABILITY ?= unstable
 
-kyma: $(LOCALBIN) $(KYMA) ## Download kyma locally if necessary.
+kyma: $(LOCALBIN) $(KYMA) ## Download Kyma cli locally if necessary.
```
chrkl, line 213: I think we should make this a `.PHONY` target since it checks for the existence of the Kyma binary itself. Please double-check.

Makefile
```diff
@@ -141,97 +134,15 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
...
-
-.PHONY: verify-telemetry
-verify-telemetry: ## Wait for Telemetry CR to be in Ready state.
-	@hack/verify_telemetry_status.sh
-
-.PHONY: verify-kyma
-verify-kyma: ## Wait for Kyma CR to be in Ready state.
-	@hack/verify_kyma_status.sh
+.PHONY: deploy-with-lifecycle-manager
+deploy-with-lifecycle-manager: kyma kustomize ## Create a k3d cluster and deploy module with the lifecycle manager. Manager image and module image are pushed to local k3d registry
```
chrkl, line 138: I think the naming of the make targets is a bit inconsistent. We have the `deploy` target that deploys the telemetry-manager to an existing cluster. This target rather represents an e2e-test for the Kyma module and includes provisioning of a k3d cluster, validating, etc.. I suggest to align the naming and the scope with the `e2e-test` target. Since the `e2e-test` also deletes the k3d cluster, let's do the same here. Also align the `##@` comments (they are used when calling `make  help`)

chrkl (changes requested):

Makefile
shorim: The target is renamed to `e2e-deploy-module`. After discussion, we decided not to delete the k3d cluster as it will be useful for development purposes as well and not only testing. For example, if we make some changes and want to deploy the module with the lifecycle-manager to check something, then we can just execute the make target to have everything setup in a local k3d cluster

chrkl (dismissed):

README.md
```diff
@@ -103,8 +103,11 @@ make deploy
 make undeploy
 ```
 
-### Deploying `ModuleTemplate` with the Lifecycle Manager
-Check the documentation [here](./docs/deploying-module-template.md).
+- Deploy module with lifecycle manager on a k3d cluster
```
grego952, line 106:
```suggestion
- Deploy module with Lifecycle Manager on a k3d cluster
```

grego952 (approved):

chrkl (approved):

PR #163, opened 2023-05-08
rename telemetry manager to operator

rakesh-garimella:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -1,27 +1,27 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: operator
   namespace: system
   labels:
-    control-plane: controller-manager
+    control-plane: operator
     app.kubernetes.io/name: deployment
-    app.kubernetes.io/instance: controller-manager
+    app.kubernetes.io/instance: operator
     app.kubernetes.io/component: manager
     app.kubernetes.io/managed-by: kustomize
 spec:
   selector:
     matchLabels:
-      control-plane: controller-manager
+      control-plane: operator
   replicas: 1
   template:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: manager
       labels:
-        control-plane: controller-manager
+        control-plane: operator
         app.kubernetes.io/name: deployment
-        app.kubernetes.io/instance: controller-manager
+        app.kubernetes.io/instance: operator
         app.kubernetes.io/component: manager
         app.kubernetes.io/managed-by: kustomize
     spec:
@@ -76,5 +76,5 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-      serviceAccountName: controller-manager
+      serviceAccountName: operator
       terminationGracePeriodSeconds: 10


config/manager/manager.yaml
```diff
@@ -43,7 +45,7 @@ spec:
         - --trace-collector-memory-request=32Mi
         - --validating-webhook-enabled=true
         - --manager-namespace=$(MY_POD_NAMESPACE)
-        image: controller:latest
+        image: k3d-kyma-registry:5001/telemetry-manager:latest
```
skhalash, line 48: Please, revert this change

config/crd/kustomization.yaml
```diff
@@ -12,7 +12,7 @@ resources:
 patchesStrategicMerge:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
-- patches/webhook_in_logpipelines.yaml
+#- patches/webhook_in_logpipelines.yaml
```
skhalash, line 15: Could you please explain why we don't need it anymore?

config/default/kustomization.yaml
```diff
@@ -10,7 +10,7 @@ namePrefix: telemetry-
 
 # Labels to add to all resources and selectors.
 labels:
-- includeSelectors: true
```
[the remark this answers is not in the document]
rakesh-garimella: We dont want to have labelselectors included from kustomize right now. We want to have the old label selectors so that the migration is smooth and we can revert this change when we rename the `operator` to `manager` again

config/crd/kustomization.yaml
rakesh-garimella: The webhook file here is a conversion webhook which was included before by mistake. Also we dont need a conversion webhook right now so I commented it out.

config/default/kustomization.yaml
```diff
@@ -10,7 +10,7 @@ namePrefix: telemetry-
...
+- includeSelectors: false
   pairs:
     app.kubernetes.io/component: telemetry
     app.kubernetes.io/part-of: telemetry
```
chrkl, line 16: Maybe let's rename the `telemetry-controller-manager` in the lines below as well.

config/manager/manager.yaml
```diff
@@ -1,29 +1,31 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: operator
   namespace: system
   labels:
-    control-plane: controller-manager
+    control-plane: operator
```
chrkl, line 7: I guess this is overwritten by the labels from `kustomization.yaml`? If we want these changes, let's move then to that place.

config/default/kustomization.yaml
skhalash: Done

config/manager/manager.yaml
skhalash: True, I moved the top-level resource labels to `kustomization.yaml`

chrkl (approved):

nataliasitko (approved):

Issue #150, opened 2023-05-02
Migrate leftover helm resources to kustomize


a-thaler:
**Description**
As part of https://github.com/kyma-project/kyma/issues/17375 the new modular deployment approach of the telemetry manager should result in the same resource setup as the old approach. For that there are some obvious resources in the helm chart not being covered yet. These need to be copied over. To not introduce a dependency to istio now, the peerAuthentication should be removed and instead the metrics port of fleuntbit excluded from istio interception.

**Criterias**
- missing Grafana Dashboard is added to to kustomize files
- missing busola extensions are added to kustomize files
- missing networkPolicy is added to kustomize files
- add a new busola extension for the telemetry CRD similar to [serverless](https://github.com/kyma-project/kyma/blob/main/resources/serverless/templates/busola-serverless-extension.yaml) containing the link to the documentation
- cover the points above in "manager tests"
- extract the "manager tests" into its own package
- peerAuthentication for fluentbit is removed (with migration script) and replaced with a port exclusion by annotation
- removed unused cluster roles from the helm chart and clean them as part of the migration script

**Reasons**

**Attachments**

PR #169, opened 2023-05-12
Migrate helm leftovers to kustomize

hisarbalik:
**Description**

- Add istio inbound port exclusion annotation for ports 2020 and 2021 
- Add networkpolicy to kustomize files
- Add busola extensions to kustomize files
- Add otel-collector and fluent-bit dashboards to kustomize files
- Extend `manager_test` to check the existence of the new resources

**Related issue(s)**
https://github.com/kyma-project/telemetry-manager/issues/150


config/networking/networkpolicy.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: telemetry-operator-pprof-deny-ingress
```
shorim, line 4: @hisarbalik the `telemetry-` prefix will be added by Kustomize https://github.com/kyma-project/telemetry-manager/blob/main/config/default/kustomization.yaml#L9. So, the name here should be only `operator-pprof-deny-ingress`

internal/resources/fluentbit/resources_test.go
```diff
@@ -48,7 +53,7 @@ func TestMakeDaemonSet(t *testing.T) {
 	require.NotEmpty(t, daemonSet.Spec.Template.Spec.Containers[0].EnvFrom)
 	require.NotNil(t, daemonSet.Spec.Template.Spec.Containers[0].LivenessProbe, "liveness probe must be defined")
 	require.NotNil(t, daemonSet.Spec.Template.Spec.Containers[0].ReadinessProbe, "readiness probe must be defined")
-
+	require.Equal(t, daemonSet.Spec.Template.ObjectMeta.Annotations, expectedAnnotations, "annotations should contains istio port exclusion of 2020 and 2021")
```
chrkl, line 56:
```suggestion
	require.Equal(t, daemonSet.Spec.Template.ObjectMeta.Annotations, expectedAnnotations, "annotations should contain istio port exclusion of 2020 and 2021")
```

PR #193, opened 2023-05-30
bump to old otel image

rakesh-garimella:
**Description**

- The e2e tests are still flaky. May be they are caused because of the last otel image bump
-
**Related issue(s)**


skhalash:
The flakiness seems not to be caused by the image bump

PR #198, opened 2023-06-05
Optional Loki LogPipeline

shorim:
**Description**

- If the `logging-loki` service (from Kyma logging component) exists in the cluster, the operator ensures that the Loki Logpipeline is created.
- If the `logging-loki` service (from Kyma logging component) doesn't exist in the cluster, the operator ensures that the Loki Logpipeline is removed.

**Related issue(s)**

kyma-project/kyma#17549


main.go
```diff
@@ -380,6 +383,21 @@ func main() {
...
+		setupLog.Error(err, "Failed to handle Loki LogPipeline")
+		os.Exit(1)
+	}
+
+	// Ensure reconciliation of the optional Loki LogPipeline
+	go func() {
+		for range time.Tick(1 * time.Minute) {
+			if err := handleLokiLogPipeline(context.Background(), k8sClient); err != nil {
+				setupLog.Error(err, "Failed to handle Loki LogPipeline")
+				os.Exit(1)
```
chrkl, line 396: We should not quit here. An error could easily happen here, for instance if the API server is not reachable. Just log the error and retry in the next iteration.

internal/kubernetes/utils.go
```diff
@@ -248,6 +278,28 @@ func mergeMapsByPrefix(newMap map[string]string, oldMap map[string]string, prefi
 	return newMap
 }
 
+func mergeFinalizers(new *metav1.ObjectMeta, old metav1.ObjectMeta) {
+	new.SetFinalizers(mergeSlices(new.Finalizers, old.Finalizers))
+}
+
+func mergeSlices(newSlice []string, oldSlice []string) []string {
```
chrkl, line 285: Maybe we can simplify this a bit. The [controllerutil package](https://github.com/kubernetes-sigs/controller-runtime/blob/30dea34848b0e0b30690dd461a0eadb74d4869f3/pkg/controller/controllerutil/controllerutil.go#L353) already contains a helper that checks for redundant finalizers. Just use the old finalizers slice to preserve the order and add each of the new finalizers.

main.go
shorim: yes, I agree  👍🏻

main.go
```diff
@@ -380,6 +383,21 @@ func main() {
 		}()
 	}
 
+	if err := handleLokiLogPipeline(context.Background(), k8sClient); err != nil {
+		setupLog.Error(err, "Failed to handle Loki LogPipeline")
+		os.Exit(1)
+	}
+
+	// Ensure reconciliation of the optional Loki LogPipeline
+	go func() {
```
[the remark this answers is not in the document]
shorim: I tried terminating the operator and it was immediately terminated without delays

main.go
```diff
@@ -554,3 +567,27 @@ func parsePlugins(s string) []string {
 func createTelemetryReconciler(client client.Client, scheme *runtime.Scheme, eventRecorder record.EventRecorder, webhookConfig telemetry.WebhookConfig) *operatorcontrollers.TelemetryReconciler {
 	return operatorcontrollers.NewTelemetryReconciler(client, telemetry.NewReconciler(client, scheme, eventRecorder, webhookConfig))
 }
+
+func handleLokiLogPipeline(ctx context.Context, client client.Client) error {
```
chrkl, line 571: Let's add an e2e test for the pipeline creation and deletion

test/e2e/logging_test.go
```diff
@@ -61,6 +66,38 @@ var _ = Describe("Logging", func() {
...
+					var lokiLogPipeline telemetryv1alpha1.LogPipeline
+					key := types.NamespacedName{Name: "loki"}
+					g.Expect(k8sClient.Get(ctx, key, &lokiLogPipeline)).To(Succeed())
+					g.Expect(lokiLogPipeline.Status.HasCondition(telemetryv1alpha1.LogPipelineRunning)).To(BeTrue())
+				}, 2*time.Minute, interval).Should(Succeed())
+			})
+		})
+
+		It("Should delete loki logpipeline", func() {
+			By("Deleting loki service")
```
chrkl, line 86: Wrap the following lines into a function

internal/resources/lokilogpipeline/resources_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMakeLokiLogPipeline(t *testing.T) {
+	lokiLogPipeline := MakeLokiLogPipeline()
+
+	require.NotNil(t, lokiLogPipeline)
+	require.Equal(t, lokiLogPipeline.APIVersion, "telemetry.kyma-project.io/v1alpha1")
```
chrkl, line 13: require.Equal wants the expected value first, then the actual one. Does not make a difference in the test execution, but gives potentially a nicer output in the case of failures

chrkl (approved):

Issue #85, opened 2023-03-19
Improve internal service exposure in the e2e tests


skhalash:
**Description**

Come up with a better solution to expose internal service in the e2e tests. It must work on both Gardener and k3d. Possible solutions:
* Use Istio gateway
* Use a LoadBalancer service
* Use k8s API server proxy

**Reasons**

In order to send, receive, and analyze traces, metrics, and logs, the e2e tests must expose internal in-cluster services. Currently, the existing tracing tests do it using `NodePort` services and k3d host port mapping. 
This approach is very easy to implement. However, it has a few major drawbacks and must be rethought:
* `NodePort` service exposure will only work on k3d and not on the public cloud (Gardener).
* One must always keep a mind map with a port map in their head. For every new internal service, a few places in the code must be updated.

**Attachments**

PR #200, opened 2023-06-06
E2e service exposure

ameteiko:
**Description**

- The E2E test suite is now available via API-server proxy.
- The code is stripped of the tricky port mapping.

**Related issue(s)**

Implements  #85


test/e2e/auth_token.go
skhalash, line 1: Hmm, I don't know if service account tokens are the best choice to auth against the API server in this case (from the outside of the cluster). I would get a bearer token or a TLS client cert from kubeconfig. It's much easier.

test/e2e/metrics_test.go
```diff
@@ -256,14 +258,23 @@ var _ = Describe("Metrics", func() {
...
-			sendMetrics(context.Background(), builder.Build(), otlpPushURL)
+			Expect(sendMetrics(context.Background(), builder.Build(), urls.otlpPush())).To(Succeed())
+			Expect(sendMetrics(context.Background(), builder.Build(), urls.otlpPushAt(1))).To(Succeed())
+
+			Eventually(func(g Gomega) {
+				resp, err := httpsAuthProvider.Get(urls.mockBackendExport())
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(resp).To(HaveHTTPStatus(http.StatusOK))
+				g.Expect(resp).To(HaveHTTPBody(SatisfyAll(
+					HaveNumberOfMetrics(len(gauges)))))
```
chrkl, line 269: Since we have two backends now, we can also validate the actual metrics (`HaveMetrics`) instead of the pure number. Same for tracing.

test/e2e/auth_token.go
skhalash: Refactored the logic to fetch the auth info from the kubeconfig

test/e2e/metrics_test.go
```diff
@@ -288,113 +299,112 @@ func metricPipelineShouldStayPending(pipelineName string) {
...
+	}
 
-	// Default namespace objects.
-	otlpEndpointURL := mockBackendExternalService.OTLPEndpointURL(grpcOTLPPort)
+	return fmt.Sprintf("%s-%d", name, idx)
+}
 
-	hostSecret1 := kitk8s.NewOpaqueSecret("metric-rcv-hostname-1", defaultNamespaceName, kitk8s.WithStringData("metric-host", otlpEndpointURL))
-	metricPipeline1 := kitmetric.NewPipeline("pipeline-1", hostSecret1.SecretKeyRef("metric-host"))
+type urlProvider struct {
```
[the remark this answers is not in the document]
ameteiko: I moved the code as requested.

At first, I was about to argue that the provider is too concrete (by exposing the awareness of the URL names), but I found a tolerable place inside the `testkit` (the `mocks` package).

test/e2e/metrics_test.go
ameteiko: That is a sharp catch! I extended the validation.

chrkl (approved):

PR #202, opened 2023-06-06
Improve fluent bit config reconciliation

chrkl:
**Description**

- Include static Fluent Bit configuration into checksum calculation
- Add owner references to all created resources
- Remove unnecessary checks for changed resources
- Add missing watch to LogParser controller

**Related issue(s)**

Preparations to fix https://github.com/kyma-project/kyma/issues/17611

### DIFF
--- a/internal/reconciler/logparser/sync_test.go
+++ b/internal/reconciler/logparser/sync_test.go
@@ -61,4 +61,5 @@ Format regex`},
 	expectedCMData := "[PARSER]\n    Name foo\n    Format regex\n\n"
 	require.Contains(t, cm.Data[parsersConfigMapKey], expectedCMData)
 	require.Len(t, cm.OwnerReferences, 1)
+	require.Equal(t, lp.Name, cm.OwnerReferences[0].Name)
 }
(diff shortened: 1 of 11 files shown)


internal/reconciler/logparser/sync_test.go
```diff
@@ -60,4 +60,5 @@ Format regex`},
 	require.NoError(t, err)
 	expectedCMData := "[PARSER]\n    Name foo\n    Format regex\n\n"
 	require.Contains(t, cm.Data[parsersConfigMapKey], expectedCMData)
+	require.Len(t, cm.OwnerReferences, 1)
```
skhalash, line 63: Let's maybe test who is the owner?

skhalash (changes requested):
Looks good 👍🏻  One question what is with the finalizers? We don't need to add them anymore, but we should cleanup if they are set?

chrkl:
> Looks good 👍🏻 One question what is with the finalizers? We don't need to add them anymore, but we should cleanup if they are set?

We cannot get rid of the finalizer yet since it is still required to clean up the pipeline specific ConfigMap keys on deletion. Adding the OwnerReferences has the purpose to reconcile the pipeline if the ConfigMap is touched by the user and also ensures that no empty ConfigMap is left after all pipelines are deleted.

skhalash (approved):

PR #208, opened 2023-06-09
add pprof extension to metric-gateway and trace-collector

dennis-ge:
**Description**

- adds the pprof extension to the metric-gateway and trace-collector
- adds dedicated networkpolicies for the metric-gateway and trace-collector
- fixes the existing networkpolicy of the operator by setting the proper labels

**Related issue(s)**

Fixes https://github.com/kyma-project/kyma/issues/17163


config/networking/networkpolicy-operator.yaml
```diff
@@ -20,4 +20,4 @@ spec:
         - protocol: TCP
           port: 8081
         - protocol: TCP
-          port: 9443
\ No newline at end of file
+          port: 9443
```
dennis-ge, line 23: the podSelector matches for `app.kubernetes.io/name: fluent-bit`. This looks wrong. Can anybody confirm this?

config/networking/networkpolicy-trace-collector.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: trace-collector-pprof-deny-ingress
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: telemetry-trace-gateway
```
a-thaler, line 8: It is probably called "collector" still

internal/otelcollector/config/builder/config_builder_test.go
```diff
@@ -98,8 +98,21 @@ func TestMakeExtensionConfig(t *testing.T) {
 		HealthCheck: config.EndpointConfig{
 			Endpoint: "${MY_POD_IP}:13133",
 		},
+		Pprof: config.EndpointConfig{
+			Endpoint: "127.0.0.1:1777",
```
chrkl, line 102: Maybe we also expose this on `${MY_POD_IP}`? This would allow a user to open it up by adding another policy and then using tools like Parca.

internal/reconciler/tracepipeline/reconciler.go
```diff
@@ -191,5 +191,14 @@ func (r *Reconciler) doReconcile(ctx context.Context, pipeline *telemetryv1alpha
 		return fmt.Errorf("failed to create otel collector metrics service: %w", err)
 	}
 
+	networkPolicyPorts := makeNetworkPolicyPorts()
+	networkPolicy := collectorresources.MakeNetworkPolicy(r.config, networkPolicyPorts)
+	if err = controllerutil.SetOwnerReference(pipeline, networkPolicy, r.Scheme()); err != nil {
```
chrkl, line 196: Add a watch for the resource during the controller setup in `tracepipeline_controller.go` to also have reconciliations triggered by modifications on the owned object. Same for the MetricPipeline.

test/e2e/metrics_test.go
```diff
@@ -378,3 +398,11 @@ func suffixize(name string, idx int) string {
 
 	return fmt.Sprintf("%s-%d", name, idx)
 }
+
+func getPodNameByLabel(namespace, labelKey, labelValue string) string {
```
chrkl, line 402: I don't like that this helper is defined in this file, but shared between the tracing and metric test. Also the assumption that it has only one pod is very specific. I'm wondering if we should rather move fetching the pod into the `Eventually` block of the test cases and improve the robustness by that while accepting duplicating these few lines of code.

chrkl (approved):

Issue #259, opened 2023-07-12
Dedicated kustomize profile used in release creation for having development features excluded from release


a-thaler:
**Description**
The metrics features is still under heavy development and should not be shipped as part of the release artifacts. To exclude dedicated features (behind a feature flag), a dedicated kustomize profile might be set using other feature toggle defaults and potentially skip resources.

**Criterias**
- Using the release artifacts of the Github release page will not activate the metrics feature and also hide the new API
- There is a new release based on that changes

**Reasons**
To not expose experimental features to the end users in production

**Attachments**

PR #265, opened 2023-07-16
Create Kustomize development variant

shorim:
**Description**

- Create Kustomize development variant to allow excluding development features from releases. Default variant will be treated as a production variant since `create module` command from the Kyma cli uses the default variant for building manifests: https://github.com/kyma-project/cli/blob/main/pkg/module/kubebuilder/project.go#L22
- Currently, `Metricpipeline` CRD will only be included in the development variant and the metrics feature will be disabled in the default variant by setting the `enable-metrics` flag to false.
- `deploy` and `undeploy` targets will be using the development variant when building the manifests. Thus, e2e-test and upgrade-test will run against the development variant.

**Related issue(s)**

#259

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,11 +168,11 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/development | kubectl apply -f -
 
 .PHONY: undeploy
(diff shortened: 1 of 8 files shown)


chrkl:
Let's keep the MetricPipeline CRD in config/crd/ and include it from that directory to the development configuration. Moving it will break the controller-gen flow.

shorim:
> Let's keep the MetricPipeline CRD in config/crd/ and include it from that directory to the development configuration. Moving it will break the controller-gen flow.

Unfortunately, this seems to be not possible. In a Kustomization file, we can only reference bases (other kustomizations) outside of the directory in which the Kustomization file is located, but in order to reference resource files, they need to be in the same directory or a subdirectory of where the kustomization.yml file is located.

https://stackoverflow.com/questions/60298445/kubernetes-kustomize-command-giving-error-when-we-specify-base-manifest-files-in
https://github.com/kubernetes-sigs/kustomize/pull/700

chrkl (dismissed):

IwonaLanger (dismissed):

chrkl (approved):

IwonaLanger (approved):

PR #266, opened 2023-07-17
Introduce make targets to run e2e tests per signal

chrkl:
**Description**

Adding more test-cases to the e2e-test suite make the test runs to take more time and also increased flakiness. This PR adds make targets to run the tests for a specific signal with the goal to improve the development workflow and have separate pipelines in future.

- Introduce make targets to run e2e tests per signal

**Related issue(s)**

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -7,5 +7,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: k3d-kyma-registry:5000/telemetry-manager
-  newTag: latest
+  newName: europe-docker.pkg.dev/kyma-project/prod/telemetry-manager
+  newTag: v20230622-dd7e6d85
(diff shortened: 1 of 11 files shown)


config/manager/kustomization.yaml
```diff
@@ -7,5 +7,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: europe-docker.pkg.dev/kyma-project/prod/telemetry-manager
-  newTag: v20230622-dd7e6d85
+  newName: k3d-kyma-registry:5000/telemetry-manager
+  newTag: latest
```
skhalash, line 11: Please, revert this change

skhalash (changes requested):

config/manager/kustomization.yaml
chrkl: Done

skhalash (approved):

PR #282, opened 2023-07-21
Simplify otelcollector config pkg structure

skhalash:
## Description

- Simplify otelcollector config pkg structure by flattening unnecessary subdirectories

## Traceability

It's neither changing the existing functionality nor adding a new one.

## Testability

Just renaming the packages, no new tests are needed, a build check is enough

## Codebase

- [ ] My code follows the style guidelines of this project.
- [x] The code was planned and designed following the defined architecture and the separation of concerns.
- [ ] The code has sufficient comments, particularly for all hard-to-understand areas.
- [x] This PR adds value and shows no feature creep.
- [ ] I have augmented the test suite that proves my fix is effective or that my feature works.


internal/otelcollector/config/config_base.go
```diff
@@ -1,4 +1,4 @@
-package common
+package config
 
 type BaseConfig struct {
```
ameteiko, line 3: I wonder if we can drop the **Config** suffix in this type and make the name just **Base**.

It will be used as `config.Base` in the downstream packages instead of `config.BaseConfig`.

internal/otelcollector/config/config_base_processors.go
```diff
@@ -1,4 +1,4 @@
-package common
+package config
 
 type BaseProcessorsConfig struct {
```
ameteiko, line 3: I also would prefer this type to be BaseProcessor. (`config.BaseProcessor` vs `config.BaseProcessorConfig`).

So the idea is to assume that all the types inside the `config` package are configs and to strip the Config suffix for them. What do you think?

skhalash: Yeah, I like the idea

ameteiko (approved):

PR #468, opened 2023-10-13
feat: Adjust filter to handle istio spans based on OTEL tracer

a-thaler:
## Description

With the switch to OTEL as tracer in istio (https://github.com/kyma-project/kyma/issues/16589), the OperationName span attribute will not be available anymore for trace filtering, also the service.name attribute will get adjusted. This PR will make the filter tolerant to the change by:
- adding an additional check for ingress/egress in the span name
- not relying on service-name for now but using k8s.namespace.name attribute and istio.canonical_service. We anyway filter only istio spans, so we can rely on istio attributes

Also it is redesigning the filters in more general by doing these changes:
- Instead of "POST Ingress to kyma grafana" it will filter "all spans of kyma grafana" as also GET requests and calls to prometheus should be filtered
- Instead of "POST Ingress to kyma auth proxy" it will filter "all spans of kyma auth proxy" as also GET requests should be filtered
- the "frontend-metrics" are covered by the above filters already. Filter "frontend-metrics" in general on the ingressgateway is dangerous, it might be emmitted by any client, so lets not filter it
- the "/metrics" filter got changed to filter all spans from kyma prometheus/vmAgent or metric-agent, nothing dedicated to /metrics anymore
- filter of calls to our services are extended to include all kind of http methods
- added new filters for ingress and egress calls of all telemetry components

## Related Issues and Documents

Closes:

Related issues:
https://github.com/kyma-project/kyma/issues/16589

## Traceability

- [ ] The PR is linked to a GitHub Issue.
- [ ] New features have a milestone label set.
- [ ] The corresponding GitHub Issue has a respective `area` label.

## Testability

The feature is unit-tested:

- [x] Yes.
- [ ] No, because unit tests are not needed.
- [ ] No, because of ...

The feature is e2e-tested:

- [x] Yes.
- [ ] No, because e2e-tests are not needed.
- [ ] No, because of ...

Tests conducted for the PR:

## Codebase

- [ ] My code follows the [Effective Go](https://go.dev/doc/effective_go) style guidelines.
- [ ] The code was planned and designed following the defined architecture and the separation of concerns.
- [ ] The code has sufficient comments, particularly for all hard-to-understand areas.
- [ ] This PR adds value and shows no feature creep.
- [ ] I have augmented the test suite that proves my fix is effective or that my feature works.


internal/otelcollector/config/trace/gateway/processors.go
```diff
@@ -82,18 +79,63 @@ func makeProcessorsConfig() Processors {
...
+	userAgentIsVMScrapeAgent           = "attributes[\"user_agent\"] == \"vm_promscrape\""
+	userAgentIsPrometheus              = "IsMatch(attributes[\"user_agent\"], \"Prometheus\\\\/.*\") == true"
+	userAgentIsKymaOtelCol             = "IsMatch(attributes[\"user_agent\"], \"kyma-otelcol\\\\/.*\") == true"
+	urlIsIstioHealthz                  = "IsMatch(attributes[\"http.url\"], \"https:\\\\/\\\\/healthz\\\\..+\\\\/healthz\\\\/ready\") == true"
+	urlIsTelemetryTraceService         = "IsMatch(attributes[\"http.url\"], \"http(s)?:\\\\/\\\\/telemetry-otlp-traces\\\\.kyma-system(\\\\..*)?:(4317|4318).*\") == true"
+	urlIsTelemetryTraceInternalService = "IsMatch(attributes[\"http.url\"], \"http(s)?:\\\\/\\\\/telemetry-trace-collector-internal\\\\.kyma-system(\\\\..*)?:(55678).*\") == true"
+	urlIsTelemetryMetricService        = "IsMatch(attributes[\"http.url\"], \"http(s)?:\\\\/\\\\/telemetry-otlp-metrics\\\\.kyma-system(\\\\..*)?:(4317|4318).*\") == true"
+	and                                = " and "
+
+	toFromKymaGrafana            = componentIsProxy + and + namespacesIsKymaSystem + and + istioNameIsGrafana
```
skhalash, line 106: Let's further simplify it by adding the following functions:
```go
func spanAttributeEquals(key, value string) string {
	return "attributes[\"" + key + "\"] == \"" + value + "\""
}

func resourceAttributeEquals(key, value string) string {
	return "resource.attributes[\"" + key + "\"] == \"" + value + "\""
}

func urlMatches(pattern string) string {
	return "IsMatch(attributes[\"http.url\"], \"" + pattern + "\") == true"
}

func userAgentMatches(pattern string) string {
	return "IsMatch(attributes[\"user_agent\"], \"" + pattern + "\") == true"
}

func joinWithAnd(parts ...string) string {
	return strings.Join(parts, " and ")
}
```

skhalash (changes requested):

internal/otelcollector/config/trace/gateway/processors.go
a-thaler: done

skhalash (approved):

Issue #670, opened 2023-12-22
Certificate with incorrect line breaks can get fluentbit in a crashloop


a-thaler:
**Description**

A typical scenario is to copy the secret for the backend configuration. Especially for the PEM bodies, you easily can do mistakes with the line breaks. As PEMs are a well-defined format, the manager should try to auto-correct typical mistakes like de-doubling backslashes.

The concrete problem here is a cert file where by mistake the line breaks are escaped (by having double backslashes). Such mistake will bring fluentbit in a crashloop with following error message:

```
[2023/12/22 15:33:53] [error] [tls] crt_file '/fluent-bit/etc/output-tls-config/sap-cloud-logging-application-logs-cert.crt' 336445449: error:0909006C:PEM routines:get_name:no start line
[2023/12/22 15:33:53] [error] [tls] could not create TLS backend
[2023/12/22 15:33:53] [error] [output http.1] error initializing TLS context
[2023/12/22 15:33:53] [error] [engine] output initialization failed
[2023/12/22 15:33:53] [error] [lib] backend failed
```

**Expected result**
It will be great if such mistakes can be self-healed and the setup will just work

**Actual result**
Currently you will need to figure out the problem and fix it manually in the setup

**Criterias**
- A cert/ca/key having escaped line breaks will lead to a functional setup

**Release Notes**

```
Certificates or keys for mTLS communication to the log backend will be sanitized with line break problems
```

rakesh-garimella:
Now we sanitize the certificates passed in the secret and remove extra escape characters.

rakesh-garimella:
Fix for release 1.6.1 https://github.com/kyma-project/telemetry-manager/pull/686

PR #675, opened 2023-12-28
fix: Validate and sanitize certificate for logpipeline

rakesh-garimella:
## Description

- Sometimes there are extra escape characters in the tls secret and key. In such case identify it and fix it by removing extra escape characters.
- Validate certificate for trace/metric pipelines as well

- https://github.com/kyma-project/telemetry-manager/issues/670

## Traceability
- [ ] New features have a milestone set.
- [x] The corresponding GitHub issue has a respective `area` and `kind` label.
- [x] The feature is unit-tested
- [ ] The feature is e2e-tested


internal/secretref/secret_ref.go
```diff
@@ -28,6 +30,19 @@ func ReferencesNonExistentSecret(ctx context.Context, client client.Reader, gett
 	return false
 }
 
+func ValidateAndSanitizeTLSSecret(tlsCert, tlsKey []byte) ([]byte, []byte) {
+	_, err := tls.X509KeyPair(tlsCert, tlsKey)
+	if err != nil {
+		if noPemDataFoundError(err) {
```
chrkl, line 36: I would do the replacement in the case of any error. Relying on the message is not very robust and this replacement should usually  not cause any harm.

internal/secretref/secret_ref.go
```diff
@@ -28,6 +30,19 @@ func ReferencesNonExistentSecret(ctx context.Context, client client.Reader, gett
 	return false
 }
 
+func ValidateAndSanitizeTLSSecret(tlsCert, tlsKey []byte) ([]byte, []byte) {
```
chrkl, line 33: Not related to secret references. Let's move this to the package where it's used.

internal/secretref/secret_ref.go
```diff
@@ -28,6 +30,29 @@ func ReferencesNonExistentSecret(ctx context.Context, client client.Reader, gett
 	return false
 }
 
+func SanitizeTlSValueOrSecret(secretData map[string][]byte, tlsKeyVariable, tlsCertVariable string) map[string][]byte {
```
chrkl, line 33: I think we can improve multiple things about this function:
1. Map arguments are references in Go, so no need to return the same map
2. Abou the naming
  a. I cannot distinguish by the name what is the difference between this and the `validateAndSanitizeTLSSecret` helper.
  b. I do not understand the `Or` in the name. This function handles only one case
  c. Replace TlS by TLS in the name
3. I still think this function should not be part of the secretref package. It works with a map, but does not have anything to do with a Secret(-reference)

internal/reconciler/logpipeline/sync.go
```diff
@@ -197,6 +199,12 @@ func (s *syncer) syncTLSConfigSecret(ctx context.Context, logPipelines []telemet
 		if err = controllerutil.SetOwnerReference(&logPipelines[i], &newSecret, s.Scheme()); err != nil {
 			return fmt.Errorf("unable to set owner reference for tls config secret: %w", err)
 		}
+
+		if tls.Cert.IsDefined() && tls.Key.IsDefined() {
```
chrkl, line 203: Maybe let's change the control flow here similar to the Otel case.

internal/secretref/secret_ref.go
rakesh-garimella: I removed this function and simplified the code

internal/reconciler/logpipeline/sync.go
rakesh-garimella: done

chrkl (dismissed):

chrkl (approved):

PR #955, opened 2024-04-09
test: Telemetry FluentBit-based components error/warning logs

### LINKED ISSUE #844: Test for error logs of active components
**Description**
Several times we missed that telemetry components are having error logs, especially in the startup phase. It will be valuable to have a test case which assures that for the regular cases, no new error logs are getting produced in any component.

**Criterias**
- All pipelines are running with all features switched on
- Not a single component of the module is producing any error log
- If there is an error log, a test case should fail

TeodorSAP:
## Description

- `collectAgentLogs` overrides flag, for including agent logs
- FluentBit-based telemetry components error/warning logs e2e test
- Refactor e2e tests and internal code, using the overrides and the `builder.BuildFluentBitConfig` function

- Resolves #844 

## Traceability
- [ ] New features have a milestone set.
- [ ] The corresponding GitHub issue has a respective `area` and `kind` label.
- [ ] The feature is unit-tested
- [ ] The feature is e2e-tested


internal/fluentbit/config/builder/config_builder_test.go
```diff
@@ -180,7 +180,7 @@ func TestMergeSectionsConfigCustomOutput(t *testing.T) {
     name             tail
     alias            foo
     db               /data/flb_foo.db
-    exclude_path     /var/log/containers/telemetry-fluent-bit-*_kyma-system_fluent-bit-*.log
+    exclude_path
```
chrkl, line 183: Let's skip this if empty. Should be easily solvable by using `SectionBuilder.AddIfNotEmpty` in `createInputSection`.

test/e2e/overrides_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/kyma-project/telemetry-manager/test/testkit/verifiers"
 )
 
-var _ = Describe("Overrides", Label("telemetry"), Ordered, func() {
+var _ = Describe("Overrides", Label("telemetry-overrides"), Ordered, func() {
```
chrkl, line 23: Was this change intentional?

test/e2e/telemetry_logs_analysis_test.go
```diff
@@ -50,10 +54,17 @@ var _ = Describe("Telemetry Components Error/Warning Logs Analysis", Label("tele
...
 
+		// logs overrides (include agent logs)
+		overrides := kitk8s.NewOverrides().WithPaused(false).WithCollectAgentLogs(true)
+		objs = append(objs, overrides.K8sObject())
+
 		// backends
 		otelCollectorLogBackend := backend.New(otelCollectorLogBackendName, mockNs, backend.SignalTypeLogs)
 		objs = append(objs, otelCollectorLogBackend.K8sObjects()...)
 		otelCollectorLogTelemetryExportURL = otelCollectorLogBackend.TelemetryExportURL(proxyClient)
+		fluentBitLogBackend := backend.New(fluentBitLogBackendName, mockNs, backend.SignalTypeLogs)
```
chrkl, line 65: Why do we need two log backends?

test/e2e/telemetry_logs_analysis_test.go
```diff
@@ -152,12 +168,22 @@ var _ = Describe("Telemetry Components Error/Warning Logs Analysis", Label("tele
...
-		// TODO: configmap: FLuentBit, exclude_path (excluding self logs)
-		// telemetry-manager/blob/test/check-error-logs/internal/fluentbit/config/builder/input.go#L15-L16
+		It("Should not have any ERROR/WARNING logs in the FluentBit containers", func() {
+			Consistently(func(g Gomega) {
+				resp, err := proxyClient.Get(fluentBitLogTelemetryExportURL)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(resp).To(HaveHTTPStatus(http.StatusOK))
+				g.Expect(resp).To(HaveHTTPBody(
+					Not(ContainLd(ContainLogRecord(SatisfyAll(
+						WithPodName(ContainSubstring("telemetry-")),
```
chrkl, line 181: Since you are searching for the same prefix as before, you just repeat the same test with the new mock backend. I suggest to be more specific about the prefix. Then you can also have specific whitelists per signal.

internal/fluentbit/config/builder/config_builder.go
```diff
@@ -15,8 +15,13 @@ type PipelineDefaults struct {
 	FsBufferLimit     string
 }
 
+type BuilderConfig struct {
```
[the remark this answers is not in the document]
TeodorSAP: Here it was on purpose. Since we thought we reffer to the configuration of the FluentBit builder, not the configuration of the config_builder itself. But if you still think using Config is better, I can change it, no problem.

test/e2e/overrides_test.go
TeodorSAP: No, I think it was only for debugging purposes. Sorry and thank you for noticing

test/e2e/telemetry_logs_analysis_test.go
TeodorSAP: Because I have two separate test cases, and I want to test otelCollector-based and FluentBit-based components separately, so that we have a more accurate understanding of what components are failing, in case something goes wrong.

test/e2e/telemetry_logs_analysis_test.go
TeodorSAP: I will see what I can do, although there are not too many components starting with "telemetry-...". It was more a way to test all our components together. But I guess the prefix condition can even be removed since it's not filtering too much out.

internal/fluentbit/config/builder/config_builder.go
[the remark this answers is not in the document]
chrkl: Good point, let's keep it.

test/e2e/telemetry_logs_analysis_test.go
chrkl: I see that  you filter already on the container name in the LogPipeline. Then the prefix is actually not important here.  Let's keep it like this to be prepared for future changes.

Issue #1127, opened 2024-05-24
Self-Monitor is Healthy if the Pipeline is broken


skhalash:
**Description**

TelemetryFlowHealthy will have the status True only if ConfigurationGenerated has the status False. If the pipeline specification is not applied to the OTel Collector (Fluent Bit) configuration, we cannot consider it healthy, as the pipeline isn't running. Currently, we lack a check for this, meaning we cannot correlate any pipeline-related alerts and assert that it is healthy.

Also, the situation has to be covered with an E2E test.

**Expected result**

TelemetryFlowHealthy should be either `False` or `Unknown`

**Actual result**

TelemetryFlowHealthy has the status `True` if `ConfigurationGenerated` has the status `False`.

**Steps to reproduce**

**Troubleshooting**

**Release Notes**

```

```

PR #1144, opened 2024-06-03
fix: Set TelemetryFlowHealthy condition to False if the pipeline is non-reconcilable 

shorim:
## Description

- If the pipeline is non-reconcilable (`ConfigurationGenerated ` condition is set `False`), then the `TelemetryFlowHealthy` condition should also be set to `False` because the telemetry data will not be delivered to the backend.
- For all the unit tests and e2e tests, check that the `TelemetryFlowHealthy` condition is set to `False` in all the scenarios for which the `ConfigurationGenerated ` condition is checked for being `False`
- For the MetricPipeline, use the reason `GatewayConfigured` instead of `AgentGatewayConfigured` in the `ConfigurationGenerated ` condition because there will be no agents in case that all of the inputs are disabled
- Improve the message for the `ConfigurationGenerated ` condition when it is `True` to make it more descriptive

- #1127

## Traceability
- [x] The PR has a milestone set.


skhalash (approved):
Looks very good!

NHingerl (approved):

PR #1145, opened 2024-06-04
fix: Add CA validation to certificate validation logic

### LINKED ISSUE #1065: Certificate validation does not take CA cert into account
**Description**

We've implemented a new feature to address client TLS certificate issues within our pipeline statuses. This enhancement can be tracked through the following GitHub issues: [Issue 930](https://github.com/kyma-project/telemetry-manager/issues/930) and [Issue 653](https://github.com/kyma-project/telemetry-manager/issues/653).

In our validation process, we currently verify both the client certificate and its corresponding private key. However, it has come to our attention that the CA certificate validation is not being adequately performed.

**Expected result**

Any invalid CA certificates will be properly identified and reflected in the pipeline status. The pipeline configuration must not be rendered.

**Actual result**

Our TLS validation process currently overlooks invalid CA certificates.

**Steps to reproduce**

**Troubleshooting**

**Release Notes**

```

```

TeodorSAP:
## Description

- Add Ca cert validation to certificate validation logic

- Resolves #1065

## Traceability
- [x] The PR has a milestone set.


TeodorSAP:
Checked `dev`, `prod`, and `stage` clusters for CAs, in order to see of the current implementation would break any existing logic.

The following commands were executed
```bash
kubectl get logpipelines -A -ojsonpath="{ range .items[*] }{ @.spec.output.http.tls }{ '\n' }{ end }" | grep '"ca"' | wc -l

kubectl get metricpipelines -A -ojsonpath="{ range .items[*] }{ @.spec.output.otlp.tls }{ '\n' }{ end }" | grep '"ca"' | wc -l

kubectl get traceppipelines -A -ojsonpath="{ range .items[*] }{ @.spec.output.otlp.tls }{ '\n' }{ end }" | grep '"ca"' | wc -l
```

[TODO] However, no CAs were found.

internal/tlscert/tls_cert_validator.go
```diff
@@ -96,34 +114,47 @@ func (v *Validator) ValidateCertificate(ctx context.Context, cert, key *telemetr
...
+	}
+
+	// Validate the certificate-key pair
 	_, err = tls.X509KeyPair(sanitizedCert, sanitizedKey)
 	if err != nil {
 		return ErrInvalidCertificateKeyPair
 	}
 
+	// Validate the CA
+	if err = validateCA(parsedCert, parsedCA); err != nil {
```
chrkl, line 130: Should we also check for CA expiration?

internal/testutils/tls_gen.go
```diff
@@ -72,9 +77,13 @@ func (c *CertBuilder) Build() (*ServerCerts, *ClientCerts, error) {
 	}
 
 	caCertPem := bytes.Buffer{}
-	err = pem.Encode(&caCertPem, &pem.Block{Type: "CERTIFICATE", Bytes: caCertBytes})
-	if err != nil {
-		return nil, nil, err
+	if !c.caInvalid {
+		err = pem.Encode(&caCertPem, &pem.Block{Type: "CERTIFICATE", Bytes: caCertBytes})
```
chrkl, line 81: Please do some investigations if there could be other possible headers for a CA bundle.

internal/tlscert/tls_cert_validator.go
```diff
@@ -142,6 +173,18 @@ func parsePrivateKey(keyPEM []byte) error {
 	return nil
 }
 
+func validateCA(cert *x509.Certificate, ca *x509.Certificate) error {
```
chrkl, line 176: What happens if the CA bundle contains a chain of root CA and intermediate CA? The check for a matching issuer below might fail in this case as well.

internal/tlscert/tls_cert_validator.go
```diff
@@ -142,6 +173,18 @@ func parsePrivateKey(keyPEM []byte) error {
...
+	if !ca.IsCA {
+		return ErrCertIsNotCA
+	}
+
+	if cert.Issuer.String() != ca.Subject.String() {
```
chrkl, line 181: This check might be wrong in some situations. This CA is used to sign the server certificate, but the configured certificate of a pipeline is the client certificate. Both certificates can be signed by the same CA, but this is not mandatory.

TeodorSAP: I think I understand now. My intuition was that the CA is always the one that signs the client certificate.

internal/tlscert/tls_cert_validator.go
TeodorSAP: I will remove the `Issuer == Subject` check

internal/tlscert/tls_cert_validator.go
TeodorSAP: Will do, thank you!

NHingerl (dismissed):

internal/testutils/tls_gen.go
TeodorSAP: According to this RFC, you can expect the following labels to be within the BEGIN header and END footer:

CERTIFICATE [[RFC5280]](https://www.rfc-editor.org/rfc/rfc5280)
X509 CRL [[RFC5280]](https://www.rfc-editor.org/rfc/rfc5280)
CERTIFICATE REQUEST [[RFC2986]](https://www.rfc-editor.org/rfc/rfc2986)
PKCS7 [[RFC2315]](https://www.rfc-editor.org/rfc/rfc2315)
CMS [[RFC5652]](https://www.rfc-editor.org/rfc/rfc5652)
PRIVATE KEY [[RFC5208]](https://www.rfc-editor.org/rfc/rfc5208) [[RFC5958]](https://www.rfc-editor.org/rfc/rfc5958)
ENCRYPTED PRIVATE KEY [[RFC5958]](https://www.rfc-editor.org/rfc/rfc5958)
ATTRIBUTE CERTIFICATE [[RFC5755]](https://www.rfc-editor.org/rfc/rfc5755)
PUBLIC KEY [[RFC5280]](https://www.rfc-editor.org/rfc/rfc5280)

> More or less, CAs are just normal certificates with some specific values, so from my perspective, we should be fine

> **_Source: https://stackoverflow.com/questions/5355046/where-is-the-pem-file-format-specified_**

internal/reconciler/tracepipeline/reconciler.go
```diff
@@ -320,7 +322,7 @@ func tlsCertValidationRequired(pipeline *telemetryv1alpha1.TracePipeline) bool {
 		return false
 	}
 
-	return otlp.TLS.Cert != nil || otlp.TLS.Key != nil
+	return otlp.TLS.Cert != nil || otlp.TLS.Key != nil || otlp.TLS.CA != nil
```
chrkl, line 325: Maybe rename also the method to something like `tlsValidationRequired`.

internal/tlscert/tls_cert_validator.go
```diff
@@ -19,25 +19,48 @@ import (
...
+	Key  *telemetryv1alpha1.ValueType
+	CA   *telemetryv1alpha1.ValueType
+}
+
 const twoWeeks = time.Hour * 24 * 7 * 2
 
+const (
+	CertExpiredErrorMessage  = "TLS certificate expired on %s"
+	CertAboutToExpireMessage = "TLS certificate is about to expire, configured certificate is valid until %s"
+	CaExpiredErrorMessage    = "TLS CA expired on %s"
```
chrkl, line 49: Let's use "CA certificate" in the message. A CA as a (legal) entity will usually not expire. Only a certificate that it uses.

NHingerl (approved):

internal/tlscert/tls_cert_validator.go
```diff
@@ -96,34 +130,64 @@ func (v *Validator) ValidateCertificate(ctx context.Context, cert, key *telemetr
...
+	now := v.now()
+	for _, ca := range parsedCAs {
+		if err = validateCA(ca, now); err != nil {
+			return err
+		}
 	}
 
-	if certExpiry.Sub(v.now()) <= twoWeeks {
-		return &CertAboutToExpireError{Expiry: certExpiry}
+	// Validate certificate expiry
```
chrkl, line 153: Let's move this for consistency also to a separate `validateCertificate` method.

chrkl: Maybe including the cert / key matching check above

chrkl (approved):

NHingerl (approved):

PR #1154, opened 2024-06-06
fix: Log load test metric fetch 

hisarbalik:
## Description

- Reduce log generator instances from 20 to 10
- Fetch pod memory, cpu, and buffer metrics with `avg_over_time` for 20 minutes time frame
- Update input and output rates metric fetching with `rate` and 20 minutes time frame  

- 

## Traceability

### DIFF
--- a/hack/load-tests/run-load-test.sh
+++ b/hack/load-tests/run-load-test.sh
@@ -327,15 +327,15 @@ function get_result_and_cleanup_metricagent() {
 
 # shellcheck disable=SC2112
 function get_result_and_cleanup_fluentbit() {
-   RECEIVED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_input_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   RECEIVED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_input_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
-   EXPORTED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_output_proc_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   EXPORTED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_output_proc_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
-   QUEUE=$(curl -fs --data-urlencode 'query=round((sum(rate(telemetry_fsbuffer_usage_bytes{service="telemetry-fluent-bit-exporter-metrics"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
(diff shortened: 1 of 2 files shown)


hack/load-tests/run-load-test.sh
```diff
@@ -327,15 +327,15 @@ function get_result_and_cleanup_metricagent() {
...
+   RECEIVED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_input_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
-   EXPORTED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_output_proc_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   EXPORTED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_output_proc_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
-   QUEUE=$(curl -fs --data-urlencode 'query=round((sum(rate(telemetry_fsbuffer_usage_bytes{service="telemetry-fluent-bit-exporter-metrics"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   QUEUE=$(curl -fs --data-urlencode 'query=round(sum(avg_over_time(telemetry_fsbuffer_usage_bytes{service="telemetry-fluent-bit-exporter-metrics"}[20m])) / 1024)' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
```
chrkl, line 334: Let's make the same adjustment for metrics and traces.

hack/load-tests/run-load-test.sh
```diff
@@ -327,15 +327,15 @@ function get_result_and_cleanup_metricagent() {
 
 # shellcheck disable=SC2112
 function get_result_and_cleanup_fluentbit() {
-   RECEIVED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_input_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   RECEIVED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_input_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
```
chrkl, line 330: Why do we not use `avg_over_time` here? The query applies the 20m only to the rate function.

hisarbalik: fluentbit_input_bytes_total and fluentbit_output_proc_bytes_total are counter type metics, if we user avg_over_time basically we will get the result as how this metrics are over time increased in average

a-thaler: then maybe apply "max_over_time" to get a scalar value and not a vector?

hack/load-tests/run-load-test.sh
hisarbalik: let us revisit metricpipeline and tracepipeline in a separate PR

hack/load-tests/run-load-test.sh
hisarbalik: max_over_time will not work properly with counter type metrics

chrkl (approved):

hack/load-tests/run-load-test.sh
a-thaler: The [xx_over_time](https://prometheus.io/docs/prometheus/latest/querying/functions/#aggregation_over_time) function expects a range vector which is the result of the [rate](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) function. Why that should not work, on what do you base that statement?

PR #1178, opened 2024-06-14
fix: Optimize telemetry load test setup

hisarbalik:
## Description

- Fix test results queries for metric pipeline, trace pipeline and log pipeline
- Load generator pods distributed evenly across cluster nodes to ensure targets are loaded evenly 

- 

## Traceability


hack/load-tests/log-fluentbit-test-setup.yaml
```diff
@@ -76,6 +76,18 @@ spec:
       labels:
         app: logs-load-generator
     spec:
+      affinity:
```
a-thaler, line 79: That will improve the situation, I agree, still it can be unequal distribution when having 10 pods on 6 nodes. Shouldn't we simply switch to a daemonset and keep the amount of pods always equal to the nodes and then increase/decrease load by config options of the pod?

chrkl, line 79: I'm wondering if we should better use a DaemonSet or at least make the number of replicas dividable by the node count.

hack/load-tests/run-load-test.sh
```diff
@@ -233,15 +233,15 @@ function cleanup() {
 
 # shellcheck disable=SC2112
 function get_result_and_cleanup_trace() {
-    RECEIVED=$(curl -fs --data-urlencode 'query=round(avg(sum(rate(otelcol_receiver_accepted_spans{service="telemetry-trace-collector-metrics"}[20m]))))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+    RECEIVED=$(curl -fs --data-urlencode 'query=round(avg(sum(rate(otelcol_receiver_accepted_spans{service="telemetry-trace-collector-metrics"}[1m]))))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
```
a-thaler, line 236: The rate method returns a value per minute, which already reflects the average increase. Then you apply sum on it which sums up all entries in this minute vector.
The next avg seems redundant as you try to aggregate something which already was aggregated, nothing left to group.
And, you still have a a value per minute in your result, so which value you will take now? So either switch back to a 20min window or call an aggregate_over_time function to get a single value

hack/load-tests/run-load-test.sh
```diff
@@ -327,11 +327,11 @@ function get_result_and_cleanup_metricagent() {
...
+   RECEIVED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_input_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
-   EXPORTED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_output_proc_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   EXPORTED=$(curl -fs --data-urlencode 'query=round((sum(rate(fluentbit_output_proc_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[5m])) / 1024))' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
-   QUEUE=$(curl -fs --data-urlencode 'query=round(sum(avg_over_time(telemetry_fsbuffer_usage_bytes{service="telemetry-fluent-bit-exporter-metrics"}[20m])) / 1024)' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   QUEUE=$(curl -fs --data-urlencode 'query=round(sum(avg_over_time(telemetry_fsbuffer_usage_bytes{service="telemetry-fluent-bit-exporter-metrics"}[5m])) / 1024)' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
```
a-thaler, line 334: You changed all windows back to 1m except for logging, here it is 5m. I don't get that inconsistency, what is the reason?

a-thaler (changes requested):

hack/load-tests/log-fluentbit-test-setup.yaml
hisarbalik: most load tests require more load generator pods, we have a 2 node setup for all test scenarios I think deployment will evenly distribute pods on two nodes

hack/load-tests/run-load-test.sh
hisarbalik: queries uodated

hack/load-tests/log-fluentbit-test-setup.yaml
hisarbalik: most load tests require more load generator pods, we have a 2 node setup for all test scenarios I think deployment will evenly distribute pods on two nodes

hack/load-tests/run-load-test.sh
hisarbalik: queries uodated

a-thaler: It is still inconsistent, here it is 5m, for everything else it is 20m, why?

hisarbalik: Fixed, in general, `avg_over_time` works better with long-range periods, with the `rate` function (we have to use rate for counters) works better with periods at least 4 times of scrape intervals, we get better results with 5 minutes, longer periods causing inconsistency with rate function

IwonaLanger (dismissed):

hack/load-tests/run-load-test.sh
```diff
@@ -360,15 +360,15 @@ function get_result_and_cleanup_fluentbit() {
 
 # shellcheck disable=SC2112
 function get_result_and_cleanup_selfmonitor() {
-   SCRAPESAMPLES=$(curl -fs --data-urlencode 'query=sum_over_time(scrape_samples_scraped{service="telemetry-self-monitor-metrics"}[5m]) / 300' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   SCRAPESAMPLES=$(curl -fs --data-urlencode 'query=sum_over_time(scrape_samples_scraped{service="telemetry-self-monitor-metrics"}[20m]) / 300' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
    SERIESCREATED=$(curl -fs --data-urlencode 'query=prometheus_tsdb_head_series{service="telemetry-self-monitor-metrics"}' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
```
chrkl, line 365: use max operator

hack/load-tests/run-load-test.sh
```diff
@@ -360,15 +360,15 @@ function get_result_and_cleanup_fluentbit() {
 
 # shellcheck disable=SC2112
 function get_result_and_cleanup_selfmonitor() {
-   SCRAPESAMPLES=$(curl -fs --data-urlencode 'query=sum_over_time(scrape_samples_scraped{service="telemetry-self-monitor-metrics"}[5m]) / 300' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
+   SCRAPESAMPLES=$(curl -fs --data-urlencode 'query=sum_over_time(scrape_samples_scraped{service="telemetry-self-monitor-metrics"}[20m]) / 300' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
```
chrkl, line 363: Check to use avg_over_time

hack/load-tests/run-load-test.sh
```diff
@@ -360,15 +360,15 @@ function get_result_and_cleanup_fluentbit() {
...
 
    SERIESCREATED=$(curl -fs --data-urlencode 'query=prometheus_tsdb_head_series{service="telemetry-self-monitor-metrics"}' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
 
    HEADSTORAGESIZE=$(curl -fs --data-urlencode 'query=prometheus_tsdb_head_chunks_storage_size_bytes{service="telemetry-self-monitor-metrics"}' localhost:9090/api/v1/query | jq -r '.data.result[] | .value[1]')
```
chrkl, line 367: Use max

hack/load-tests/run-load-test.sh
hisarbalik: we need the ingestion rate here, the formula to calculate it is  either 
ingestion_rate = targets_count * metrics_per_target / scrape_interval or
ingestion_rate = active_time_series / scrape_interval

hack/load-tests/run-load-test.sh
hisarbalik: see comment above

hack/load-tests/run-load-test.sh
hisarbalik: max added for others metrics

chrkl (approved):

a-thaler (approved):

IwonaLanger (approved):

PR #1227, opened 2024-06-28
feat: Collect loadtest results

k15r:
## Description

- filter the matrix for loadtests to only include relevant tests
- allow selection of test case using drop down
- collect all test results and automatically print then at the end of the tests
- remove hibernation configuration from shoot configuration to avoid time related test issues (machines were hibernated during tests)
- force gardener deprovisioning even if tests failed

- 

## Traceability

### DIFF
--- a/hack/shoot_gcp.yaml
+++ b/hack/shoot_gcp.yaml
@@ -2,8 +2,6 @@ apiVersion: core.gardener.cloud/v1beta1
 kind: Shoot
 metadata:
   name: ${GARDENER_CLUSTER_NAME}
-  annotations:
-    dashboard.garden.sapcloud.io/no-hibernation-schedule: 'true'
 spec:
   secretBindingName: ${GARDENER_SECRET_NAME}
   cloudProfileName: gcp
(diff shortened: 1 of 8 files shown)


hack/shoot_gcp.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: core.gardener.cloud/v1beta1
 kind: Shoot
 metadata:
   name: ${GARDENER_CLUSTER_NAME}
+  annotations:
+    dashboard.garden.sapcloud.io/no-hibernation-schedule: 'true'
```
a-thaler, line 6: If a cluster cannot get deprovisioned and just stays. Then it will be good if it hibernates till someone realizes it and will manually delete it. -> Are you sure to disable it?

k15r: hibernation was disabled all the time (on purpose but still with broken configuration). this annotation just removes a ui hint that tells you: "this cluster has no hibernation schedule enabled... are you sure about this?"

a-thaler: Then for the gardener integration jobs at least it is enabled and it was helpful in the past as it saved many for orphaned instances, why not trying to do it here as well?

hisarbalik (approved):

PR #1251, opened 2024-07-05
chore: Check for warnings/errors in telemetry-self-monitor

a-thaler:
## Description

- Refactored the log test to analyze all logs with one test setup
- Enabled the test (a wrong label avoided execution)
- Added more excludes for otel collector components
- Adjusted fluentbit verification to search for errors in the log body as it is not logging in json
- Added new test case for the self-monitor

- 

## Traceability

### DIFF
--- a/test/e2e/telemetry_logs_analysis_test.go
+++ b/test/e2e/telemetry_logs_analysis_test.go
@@ -167,7 +167,7 @@ var _ = Describe(suite.ID(), Label(suite.LabelTelemetryLogAnalysis), Ordered, fu
 			assert.DeploymentReady(ctx, k8sClient, types.NamespacedName{Namespace: namespace, Name: selfMonitorLogBackendName})
 		})
 
-		It("Should have a running agents", func() {
+		It("Should have running agents", func() {
 			assert.DaemonSetReady(ctx, k8sClient, kitkyma.MetricAgentName)
 			assert.DaemonSetReady(ctx, k8sClient, kitkyma.FluentBitDaemonSet)
 		})
--- a/test/e2e/telemetry_logs_analysis_test.go
+++ b/test/e2e/telemetry_logs_analysis_test.go
@@ -3,7 +3,6 @@
 package e2e
 
 import (
-	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


test/e2e/telemetry_logs_analysis_test.go
```diff
@@ -25,117 +24,128 @@ import (
...
-			"ERROR", "error",
-			"WARNING", "warning",
-			"WARN", "warn"}
+		traceBackendURL            string
+		metricBackendURL           string
+		logBackendURL              string
+		otelCollectorLogBackendURL string
+		fluentBitLogBackendURL     string
+		selfMonitorLogBackendURL   string
+		namespace                  = suite.ID()
```
TeodorSAP, line 45: I would still recommend separating them into 3 namespaces. You can use the `suite.IDWithSuffix("...")` function

test/e2e/telemetry_logs_analysis_test.go
```diff
@@ -148,80 +158,99 @@ var _ = Describe(suite.ID(), Label(suite.LabelTelemetryLogsAnalysis), Ordered, f
...
+
+		It("Should collect fluent-bit component logs successfully", func() {
+			assert.LogsDelivered(proxyClient, "telemetry-", fluentBitLogBackendURL)
+		})
+
+		It("Should collect self-monitor component logs successfully", func() {
+			assert.LogsDelivered(proxyClient, "telemetry-", selfMonitorLogBackendURL)
+		})
+
+		It("Should not have any error/warn logs in the otel collector component containers", func() {
```
TeodorSAP, line 209: I like that the code looks simpler and more intuitive now that everything is in the same context. But this approach will make things harder for someone who wants to check exactly what failed, since if the otel It statement fails, for example, the other two (fluentbit, selfmon) will be skipped. Maybe we can discuss this during the daily.

test/e2e/telemetry_logs_analysis_test.go
```diff
@@ -148,80 +158,99 @@ var _ = Describe(suite.ID(), Label(suite.LabelTelemetryLogsAnalysis), Ordered, f
...
+								ContainSubstring("Variable substitution using $VAR will be deprecated"),
 							),
 						)),
 					)))),
 				))
 			}, consistentlyTimeout, periodic.TelemetryInterval).Should(Succeed())
 		})
 
-		AfterAll(func() {
-			format.MaxLength = gomegaMaxLength // restore Gomega truncation
```
TeodorSAP, line 194: You need to restore gomega truncation after executing this test suite, since otherwise the other e2e tests might be influenced by it.

TeodorSAP: > **Note:** Gomega truncation was set to unlimited/disabled for this test suite, since some errors or error messages weren't properly caught by the matchers otherwise (Usually when long logs are involved).

```go
format.MaxLength = 0 // remove Gomega truncation
```

TeodorSAP (changes requested):

test/e2e/telemetry_logs_analysis_test.go
TeodorSAP: It is fine to have everything in the same context (as discussed in the daily).

test/e2e/telemetry_logs_analysis_test.go
a-thaler: You would separate the backends into 3 namespaces? I changed the approach to have one context only, so having 3 namespaces brings no benefit in my understanding? Why you think so?

test/e2e/telemetry_logs_analysis_test.go
```diff
@@ -25,117 +24,128 @@ import (
...
 
-		// metrics & traces
-		metricPipelineName = fmt.Sprintf("%s-pipeline", metricBackend.Name())
+		//pipeline
 		metricPipeline := testutils.NewMetricPipelineBuilder().
-			WithName(metricPipelineName).
-			WithPrometheusInput(true, testutils.IncludeNamespaces(otelCollectorNs)).
-			WithRuntimeInput(true, testutils.IncludeNamespaces(otelCollectorNs)).
-			WithIstioInput(true, testutils.IncludeNamespaces(otelCollectorNs)).
+			WithName(backendName).
```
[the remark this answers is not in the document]
a-thaler: It would require another set of constants (5 pieces) for these names, but they actually bring no real value. We could rename the variable here to "scenarioName" if you like. But I see no benefit in prefixing the pipelines with a "pipeline" string.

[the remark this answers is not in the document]
TeodorSAP: It was just more for uniformity's sake, since this is how we usually write the e2e tests, but I agree with you, your approach is more straight forward, you can leave it like this.

test/e2e/telemetry_logs_analysis_test.go
TeodorSAP: I was just thinking of separating the 3 scenarios, so they don't influence each other in any way. But you are right, I think with your current implementation you won't be able to properly gather the logs if you split it in namespaces. Leave it like this then.

test/e2e/telemetry_logs_analysis_test.go
TeodorSAP: You are right, sorry

TeodorSAP (approved):

Issue #1182, opened 2024-06-17
New metric input for kyma status as annotation


a-thaler:
**Description**
As part of https://github.com/kyma-project/telemetry-manager/issues/728 have an annotation for MetricPipelines ready which will activate the new receiver for collecting kyma module status. 

The annotation might be named like:
```
telemetry.kyma-project.io/kymastats: true
```

**Criterias**
- Have an annotation in place which is disabled by default and can be activated by setting the value to "true"
- If enabled, the new kymastats receiver gets enabled wrapped by the leader-election receiver and will start collecting metrics for all modules
- The receiver is either running in the agent or the gateway, decision pending on that
- The instrumentationScope is defined as for other inputs
- If one MetricPipeline is activating the new input while a second does not, the second pipeline should not export the data
- Have e2e tests in place

**Reasons**

**Attachments**

**Release Notes**

```

```

a-thaler:
A config which works for me:
```
        singleton_receiver_creator:
          auth_type: serviceAccount
          leader_election:
            lease_name: telemetry-metric-agent-kymastats
            lease_namespace: kyma-system
          receiver:
            kymastats:
              auth_type: serviceAccount
              collection_interval: 30s
              module_groups:
              - operator.kyma-project.io
              - connectivityproxy.sap.com
              - operator.sme.sap.com
```

PR #1321, opened 2024-08-04
feat: Allow collecting Telemetry module status metrics in development variant

shorim:
## Description

- Move the `KymaInputAllowed` field from the MetricPipeline controller config to the reconciler config, so that it can be passed down to the gateway config builder
- When the feature flag `--kyma-input-allowed` is true and a MetricPipeline exists with the annotation `telemetry.kyma-project.io/experimental-kyma-input` set to true, then the Telemetry module status metrics are collected by the Metric Gateway. This is done by running a [kymastatsreceiver](https://github.com/kyma-project/opentelemetry-collector-components/tree/main/receiver/kymastatsreceiver) wrapped inside a [singletonreceivercreator](https://github.com/kyma-project/opentelemetry-collector-components/tree/main/receiver/singletonreceivercreator)
- For the instrumentationScope of the Kyma status metrics, set the name to `io.kyma-project.telemetry/kyma` and set the version to the value of the module version (same as for other inputs)
- Move the transform processor for setting the instrumentationScope from the `internal/otelcollector/config/metric/agent` pkg to the `internal/otelcollector/config/metric` pkg to make it accessible for both the agent and the gateway
- Rename the e2e test label `v1beta1` to `experimental` so that it can be used for labelling tests of all experimental features
- Adjust the creation of the `FlatMetric` in the e2e testkit to be created for each metric data point and not just one instance for all the metric data points.

- Add unit and e2e tests

- #1182 

## Traceability


internal/otelcollector/config/metric/gateway/receivers.go
```diff
@@ -0,0 +1,46 @@
...
+			},
+		},
+	}
+}
+
+func makeSingletonKymaStatsReceiverCreatorConfig(gatewayNamespace string) *SingletonKymaStatsReceiverCreator {
+	return &SingletonKymaStatsReceiverCreator{
+		AuthType: "serviceAccount",
+		LeaderElection: LeaderElection{
+			LeaseName:      "telemetry-metric-gateway-kymastats",
```
skhalash, line 29: Should not we pass the name of the lease as a parameter?

internal/otelcollector/config/metric/gateway/receivers.go
```diff
@@ -0,0 +1,46 @@
...
+				HTTP: config.Endpoint{
+					Endpoint: fmt.Sprintf("${%s}:%d", config.EnvVarCurrentPodIP, ports.OTLPHTTP),
+				},
+				GRPC: config.Endpoint{
+					Endpoint: fmt.Sprintf("${%s}:%d", config.EnvVarCurrentPodIP, ports.OTLPGRPC),
+				},
+			},
+		},
+	}
+}
```
skhalash, line 23: It's a little bit odd to have a function that makes receivers config, which does not call `makeSingletonKymaStatsReceiverCreatorConfig`, defined next to it. Would it be possible to move the annotation check here?

test/e2e/metrics_kyma_input_test.go
```diff
@@ -0,0 +1,172 @@
...
+							HaveField("Name", "kyma.module.status.state"),
+							HaveField("MetricAttributes", HaveKey("state")),
+							HaveField("ResourceAttributes", HaveKeyWithValue("k8s.namespace.name", kitkyma.SystemNamespaceName)),
+							HaveField("ResourceAttributes", HaveKeyWithValue("kyma.module.name", "Telemetry")),
+							HaveField("ScopeAndVersion", HaveField("Name", metric.InstrumentationScopeKyma)),
+							HaveField("ScopeAndVersion", HaveField("Version", SatisfyAny(
+								Equal("main"),
+								MatchRegexp("[0-9]+.[0-9]+.[0-9]+"),
+							))),
+						)),
```
skhalash, line 110: Can we maybe split this into 4 consequent checks, each check including `ContainElement` to avoid deep nesting

shorim: Done

internal/otelcollector/config/metric/gateway/receivers.go
shorim: We agreed that since this is the only place so far where the lease name is used, so there is no need now to pass it as a parameter

internal/otelcollector/config/metric/gateway/receivers.go
shorim: We agreed to leave it as it is because the `makeReceiversConfig()` is used only to create the shared receivers for ALL the pipelines. Then in the config builder, we iterate over all the pipelines and add extra receivers/processor needed depending on the spec of the pipelines.

skhalash (approved):

Issue #1358, opened 2024-08-19
Module status is flaky on rare log data flow


a-thaler:
**Description**
If log data is processed only from time to time, for example when access logs are processed, the module status is reported as "No LogsDelivered" for exactly one datapoint only. 

Debugging the situation shows that ingestion and export metrics are working as expected with a very tiny time shift, however the two alert rules are combined in a bad way.

**Criterias**
- If ingestion and export is healthy, the module status is healthy again all the time, even if there is a time shift of few seconds.

**Reasons**

**Attachments**

**Release Notes**

```

```

PR #1397, opened 2024-08-28
fix: False positive NoLogsDelivered LogPipeline status reason

skhalash:
## Description

- Before the `NoLogsDelievered` status reason was calculated based on 2 alerts: `LogAgentReceiverReadLogs` and `LogAgentExporterSentLogs`. This caused some flakiness when both alert statuses were not updated simultaneously. In this PR, a new alert was introduced that is logically the same, but boolean logic is done in PromQL instead of the Go code. Moreover, the `for` alert attribute was set to 1 minute to prevent flakiness.

- https://github.com/kyma-project/telemetry-manager/issues/1358

## Traceability


internal/selfmonitor/config/fluent_bit_rule_builder.go
```diff
@@ -74,6 +71,24 @@ func (rb fluentBitRuleBuilder) bufferFullRule() Rule {
...
 
+func (rb fluentBitRuleBuilder) noLogsDeliveredRule() Rule {
+	receiverReadExpr := rate(metricFluentBitInputBytesTotal, selectService(fluentBitMetricsServiceName)).
+		sumBy(labelPipelineName).
+		greaterThan(0).
+		build()
+
+	exporterNotSentExpr := rate(metricFluentBitOutputProcBytesTotal, selectService(fluentBitMetricsServiceName)).
+		sumBy(labelPipelineName).
+		equal(0).
```
TeodorSAP, line 82: Maybe it's a stupid idea, but wouldn't it be safer to use "<="

internal/selfmonitor/prober/log_pipeline_prober_test.go
```diff
@@ -61,6 +61,25 @@ func TestLogPipelineProber(t *testing.T) {
...
 			},
 		},
+		{
+			name:         "pending alert",
+			pipelineName: "cls",
+			alerts: promv1.AlertsResult{
+				Alerts: []promv1.Alert{
+					{
+						Labels: model.LabelSet{
+							"alertname": "LogAgentBufferFull",
```
TeodorSAP, line 71: What is this test case testing? Are we using the pending state on the LogAgentBufferFull as well?

TeodorSAP: Shouldn't this test target the "NoLogsDelivered" pending state?

internal/selfmonitor/config/fluent_bit_rule_builder.go
skhalash: Rate cannot be negative

internal/selfmonitor/prober/log_pipeline_prober_test.go
skhalash: No, this is the whole idea of this change. If an alert is pending, we should not map to the condition reason, only if it's firing.

TeodorSAP: Ok, but here you are checking that if it's pending, `Healthy: true` right?

TeodorSAP (approved):

Issue #1317, opened 2024-08-02
Pushing metrics to Dynatrace produce DIMENSION_KEY_OR_VALUE_EMPTY warning


rakesh-garimella:
**Description**

Following the [example](https://kyma-project.io/#/telemetry-manager/user/integration/dynatrace/README?id=ingest-metrics) provided for pushing metrics to Dynatrace endpoint produce warning messages

```
2024-07-31T08:34:14.297Z	warn	otlphttpexporter@v0.99.0/otlp.go:358	Partial success response	{"kind": "exporter", "data_type": "metrics", "name": "otlphttp", "message": "The following issues were encountered while ingesting OTLP metrics:\nWarnings:\nDimension dropped. key: 'k8s.namespace.name' value: '' - Reason: DIMENSION_KEY_OR_VALUE_EMPTY\nDimension dropped. key: 'k8s.namespace.name' value: '' - Reason: DIMENSION_KEY_OR_VALUE_EMPTY\n", "dropped_data_points": 0}
```

Enabling debug we see following metric with empty namespace name
```
Value: 1
ResourceMetrics #47
Resource SchemaURL:
Resource attributes:
     -> k8s.namespace.name: Str()
     -> kyma.module.name: Str(APIGateway)
     -> k8s.pod.ip: Str(xx.xxx.xx.xxx)
     -> k8s.daemonset.name: Str(telemetry-metric-agent)
     -> k8s.node.name: Str(xxxxxxxxxxxxx)
     -> kyma.kubernetes_io_app_name: Str(telemetry-metric-agent)
     -> k8s.pod.name: Str(telemetry-metric-agent-ddntj)
     -> k8s.cluster.name: Str(xxxxxxxxxxxxxxxxxxx)
     -> service.name: Str(telemetry-metric-agent)
```

We did not see any warnings from the k8sattribute processor

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

**Release Notes**

```

```

a-thaler:
It seems that dynatrace has problems in handling metric (resource) attributes having an empty value. We should add a transform statement to the example to drop the (resource) attributes when empty.

rakesh-garimella:
The changes have been merged to delete resource attributes if empty. Currently there is no way of generic drop statement to find empty resource attributes dynamically and drop them. We can only drop in hard coded way.

PR #1414, opened 2024-08-30
docs: Improve  dynatrace config to drop empty attributes and resource attributes

rakesh-garimella:
## Description

- There can be a case where attributes/resource attribute keys might have empty values. In such case these keys are dropped by the dynatrace backend (accepting the metric). There is warning in such a case which can cause confusion.

Unfortunately ottl does not support generic way of finding all the keys with values so currently hardcoding the keys which can have possible issue. 

- https://github.com/kyma-project/telemetry-manager/issues/1317

## Traceability

### DIFF
--- a/docs/user/integration/dynatrace/exporter-values.yaml
+++ b/docs/user/integration/dynatrace/exporter-values.yaml
@@ -6,6 +6,7 @@ mode: "deployment"
 
 image:
   repository: "otel/opentelemetry-collector-contrib"
+  tag: "0.107.0"
 
 config:
   exporters:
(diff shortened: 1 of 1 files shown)


docs/user/integration/dynatrace/exporter-values.yaml
```diff
@@ -39,8 +40,20 @@ config:
...
+          - delete_key(resource.attributes, "k8s.cluster.name") where resource.attributes["k8s.cluster.name"] == ""
+      - context: resource
+        statements:
+          # Remove empty attributes
+          - delete_key(attributes, "k8s.node.name") where attributes["k8s.node.name"] == ""
+          - delete_key(attributes, "k8s.namespace.name") where attributes["k8s.namespace.name"] == ""
+          - delete_key(attributes, "k8s.cluster.name") where attributes["k8s.cluster.name"] == ""
 
-    # Selects only one metric for the current collector pod 
+    # Selects only one metric for the current collector pod
```
NHingerl, line 56:
```suggestion
    # Selects only one metric for the current collector Pod
```

NHingerl:
Why are tws codeowner for a yaml file tho?

docs/user/integration/dynatrace/exporter-values.yaml
```diff
@@ -39,8 +40,20 @@ config:
...
+      - context: metric
+        statements:
+          # Remove empty resource attributes
+          - delete_key(resource.attributes, "k8s.node.name") where resource.attributes["k8s.node.name"] == ""
+          - delete_key(resource.attributes, "k8s.namespace.name") where resource.attributes["k8s.namespace.name"] == ""
+          - delete_key(resource.attributes, "k8s.cluster.name") where resource.attributes["k8s.cluster.name"] == ""
+      - context: resource
+        statements:
+          # Remove empty attributes
+          - delete_key(attributes, "k8s.node.name") where attributes["k8s.node.name"] == ""
```
a-thaler, line 52: Is that really meaningful? Will it ever happen that these attributes are not resource attributes.. and then are empty?

docs/user/integration/dynatrace/exporter-values.yaml
```diff
@@ -39,8 +40,20 @@ config:
         statements:
           # convert the <histogram_name>_sum metrics to gauges.
           - convert_sum_to_gauge() where IsMatch(metric.name, ".*_sum")
+      - context: metric
+        statements:
+          # Remove empty resource attributes
+          - delete_key(resource.attributes, "k8s.node.name") where resource.attributes["k8s.node.name"] == ""
+          - delete_key(resource.attributes, "k8s.namespace.name") where resource.attributes["k8s.namespace.name"] == ""
+          - delete_key(resource.attributes, "k8s.cluster.name") where resource.attributes["k8s.cluster.name"] == ""
```
a-thaler, line 48: The cluster name cannot be empty. If it is the case, we must fix it..

docs/user/integration/dynatrace/exporter-values.yaml
rakesh-garimella: removed

docs/user/integration/dynatrace/exporter-values.yaml
rakesh-garimella: removed

PR #1433, opened 2024-09-10
feat: Increase the rsa key size

rakesh-garimella:
## Description

-  change rsa key size to 4096
- check the current key. If its not 4096 bits long then regenerate it.
- Increase the CPU time and intialdelay for livenessprobe as generating the cert with 4096 bits is CPU intensive. It might sometimes take long for the generation and during this time the liveness probe might fail causing restart of the container.

- 

## Traceability


internal/webhookcert/ca_key_length_checker.go
skhalash, line 1: Can we simplify it by calling it key_length_checker without (ca)? Actually, it has nothing to do with CA cert

internal/webhookcert/ca_cert_provider.go
```diff
@@ -75,31 +77,36 @@ func (p *caCertProviderImpl) provideCert(ctx context.Context, caSecretName types
 }
 
 func (p *caCertProviderImpl) checkCASecret(ctx context.Context, caSecret *corev1.Secret) bool {
```
skhalash, line 79: Let's add more logs to this functions. We have to distinguish between 3 different situations:
1) No CA cert found - create new.
2) CA cert expired - create new.
3) Key size mismatch - create new.

It would be nice to have 3 distinct log messages.

skhalash (changes requested):

internal/webhookcert/ca_cert_provider.go
rakesh-garimella: after discussion decided to skip this

internal/webhookcert/ca_cert_provider_test.go
```diff
@@ -21,6 +21,14 @@ func (g *mockCACertGenerator) generateCert() ([]byte, []byte, error) {
 	return g.cert, g.key, nil
 }
 
+type mockCertKeyLengthChecker struct {
+	keyLenValid bool
+}
```
skhalash, line 26:
```suggestion
type mockKeyLengthChecker struct {
	keyLenValid bool
}
```

internal/webhookcert/key_length_checker.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+	"fmt"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type caKeyLengthChecker interface {
+	checkKeyLength(ctx context.Context, keyPEM []byte) (bool, error)
+}
```
skhalash, line 12:
```suggestion
type keyLengthChecker interface {
	checkKeyLength(ctx context.Context, keyPEM []byte) (bool, error)
}
```

rakesh-garimella: fixed

internal/webhookcert/ca_cert_provider_test.go
rakesh-garimella: fixed

skhalash (approved):

Issue #1301, opened 2024-07-25
Enhance runtime input with selectors for PVC volumes


a-thaler:
**Description**
As part of https://github.com/kyma-project/telemetry-manager/issues/972 enhance the current runtime input with a new selector for volumes. It should be disabled by default.

**Criterias**
- A new selector is added to the runtime input which will emit metrics as outlined in https://github.com/kyma-project/telemetry-manager/issues/1001#issuecomment-2154182359
- The selector can be used independent of the other selectors, if this selector is enabled only, only the related metrics are exported
- The new metrics are based on the existing k8sclusterreicever being wrapped with the leader-election-receiver
- Multiple pipelines having different selector settings will result in different metrics being exported
- Assure that only metrics for volumes being attached to a PVC are exported. Other volumes should be filtered out
- Check with the Cloud Manager team how much these metrics are useful for NFS volumes and decide if they should be included or excluded
- The k8sclusterreicever is running at the agent instead of gateway as it is representing an optional input
- There is no runtime instrumentation scope processor in the gateway

**Tasks**
- Add new feature with filter for PVC volumes
- Align with cloud manager about NFC volumes
- Verify proper instrumentation scope is in use
- Enhance documentation
- Enhance E2E test for runtime input (to assure compatibility of the emmitted metrics)

**Reasons**

**Attachments**

**Release Notes**

```

```

rakesh-garimella:
The changes wrt to volume metrics has been implemented and also the k8sCluster has been moved to metric agent

PR #1486, opened 2024-10-01
feat: Add a selector for runtime volume metrics

shorim:
## Description

- Add a selector in `MetricPipeline` CRDs for enabling/disabling the collection of volume metrics
- Refactor the creation of `metric_groups` list in `kubeletstats` receiver: a metric group is now only added to the list if at least one `MetricPipeline` has the corresponding input selector enabled. This will let us avoid sending unnecessary metrics to the gateway which will be dropped by all pipelines.
- Extract the checks for the enabled inputs to the `metrics` pkg, so that it can be used in both `gateway` and `agent` packages
- In order to add the `k8s.volume.type` resource attribute to volume metrics, `extra_metadata_labels` field is added to the `kubeletstats` receiver and the permission for the agent to read the `nodes/proxy` resource is added to the `ClusterRole` of the agent
- Add `filter/drop-non-pvc-volumes-metrics` in agent to filter only for PVC volume metrics and drop metrics from other volumes types
- Add `filter/drop-runtime-volume-metrics` in gateway to drop volume metrics for pipelines in which volume metrics are disabled
- Add unit and e2e tests
- Adjust busola `ConfigMap` for `MetricPipeline`

- #1301

## Traceability
- [x] If a CRD is changed, the corresponding Busola ConfigMap has been adjusted.


internal/otelcollector/config/metric/agent/config_builder.go
```diff
@@ -128,3 +187,15 @@ func makePipelinesConfig(inputs inputSources) config.Pipelines {
...
 }
+
+func makeRuntimePipelineProcessorsIDs(runtimeResources runtimeResourcesEnabled) []string {
+	processors := []string{"memory_limiter"}
+
+	if runtimeResources.volume {
+		processors = append(processors, "filter/drop-non-pvc-volumes-metrics")
+	}
+
+	processors = append(processors, "resource/delete-service-name", "transform/set-instrumentation-scope-runtime", "transform/insert-skip-enrichment-attribute", "batch")
```
k15r, line 198: is the order important? If not it would imho read better like this:
```suggestion
	processors := []string{"memory_limiter"
	"resource/delete-service-name", 
	"transform/set-instrumentation-scope-runtime", 
	"transform/insert-skip-enrichment-attribute", 
	"batch",
	}

	if runtimeResources.volume {
		processors = append(processors, "filter/drop-non-pvc-volumes-metrics")
	}
```

internal/otelcollector/config/metric/agent/processors_test.go
```diff
@@ -115,4 +115,19 @@ func TestProcessors(t *testing.T) {
...
+
+	t.Run("drop non-PVC volumes metrics processor", func(t *testing.T) {
+		collectorConfig := sut.Build([]telemetryv1alpha1.MetricPipeline{
+			testutils.NewMetricPipelineBuilder().WithRuntimeInput(true).WithRuntimeInputVolumeMetrics(true).Build(),
+		}, BuildOptions{})
+
+		expectedDropNonPVCVolumesMetricsProcessor := FilterProcessor{
+			Metrics: FilterProcessorMetrics{
+				Metric: []string{
+					"resource.attributes[\"k8s.volume.name\"] != nil and resource.attributes[\"k8s.volume.type\"] != \"persistentVolumeClaim\"",
```
k15r, line 127: lets use raw strings and get rid of those escapes
```suggestion
					`resource.attributes["k8s.volume.name"] != nil and resource.attributes["k8s.volume.type"] != "persistentVolumeClaim"`,
```

internal/otelcollector/config/metric/agent/receivers_test.go
```diff
@@ -55,6 +55,169 @@ func TestReceivers(t *testing.T) {
...
+				K8sNodeCPUUtilization:        MetricConfig{Enabled: false},
+				K8sNodeCPUTime:               MetricConfig{Enabled: false},
+				K8sNodeMemoryMajorPageFaults: MetricConfig{Enabled: false},
+				K8sNodeMemoryPageFaults:      MetricConfig{Enabled: false},
+				K8sNodeNetworkIO:             MetricConfig{Enabled: false},
+				K8sNodeNetworkErrors:         MetricConfig{Enabled: false},
+			},
+			ExtraMetadataLabels: []string{
+				"k8s.volume.type",
+			},
```
k15r, line 220: This would be a good use for a table test as the difference in input and expected output for each of those tests is quite hard to spot

internal/otelcollector/config/metric/gateway/processors_test.go
```diff
@@ -334,6 +334,29 @@ func TestProcessors(t *testing.T) {
...
+					Build(),
+			},
+			BuildOptions{},
+		)
+		require.NoError(t, err)
+
+		expectedDropRuntimeVolumeMetricsProcessor := FilterProcessor{
+			Metrics: FilterProcessorMetrics{
+				Metric: []string{
+					"instrumentation_scope.name == \"io.kyma-project.telemetry/runtime\" and IsMatch(name, \"^k8s.volume.*\")",
```
k15r, line 353: raw strings btw

internal/otelcollector/config/metric/input_enabled.go
```diff
@@ -0,0 +1,67 @@
...
+
+func IsRuntimePodInputEnabled(input telemetryv1alpha1.MetricPipelineInput) bool {
+	// Define first isRuntimePodMetricsDisabled to ensure that the runtime pod metrics will be enabled by default
+	// in case any of the fields (Resources, Pod or Enabled) is nil
+	isRuntimePodMetricsDisabled := input.Runtime.Resources != nil &&
+		input.Runtime.Resources.Pod != nil &&
+		input.Runtime.Resources.Pod.Enabled != nil &&
+		!*input.Runtime.Resources.Pod.Enabled
+
+	return !isRuntimePodMetricsDisabled
```
k15r, line 37:
```suggestion
	if input.Runtime.Resources == nil || input.Runtime.Resources.Pod == nil || input.Runtime.Resources.Pod.Enabled == nil {
		return true
	}
	return *input.Runtime.Resources.Pod.Enabled
```

internal/otelcollector/config/metric/input_enabled.go
```diff
@@ -0,0 +1,67 @@
...
+
+func IsRuntimeContainerInputEnabled(input telemetryv1alpha1.MetricPipelineInput) bool {
+	// Define first isRuntimeContainerMetricsDisabled to ensure that the runtime container metrics will be enabled by default
+	// in case any of the fields (Resources, Pod or Enabled) is nil
+	isRuntimeContainerMetricsDisabled := input.Runtime.Resources != nil &&
+		input.Runtime.Resources.Container != nil &&
+		input.Runtime.Resources.Container.Enabled != nil &&
+		!*input.Runtime.Resources.Container.Enabled
+
+	return !isRuntimeContainerMetricsDisabled
```
k15r, line 48:
```suggestion
	if input.Runtime.Resources == nil || input.Runtime.Resources.Container == nil || input.Runtime.Resources.Container.Enabled == nil {
		return true
	}
	return *input.Runtime.Resources.Container.Enabled
```

internal/otelcollector/config/metric/agent/config_builder.go
shorim: It is better in terms of performance to drop the unwanted metrics in the beginning to avoid doing all the other processing for metrics which will be dropped at the end.

But I'm not strongly against doing your suggestion for better readability. What do you think?

internal/otelcollector/config/metric/input_enabled.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func IsRuntimeNodeInputEnabled(input telemetryv1alpha1.MetricPipelineInput) bool {
+	// Runtime node metrics are disabled by default
+	// If any of the fields (Resources, Node or Enabled) is nil, the node metrics will be disabled
+	return input.Runtime.Resources != nil &&
+		input.Runtime.Resources.Node != nil &&
+		input.Runtime.Resources.Node.Enabled != nil &&
+		*input.Runtime.Resources.Node.Enabled
```
k15r, line 57:
```suggestion
	if input.Runtime.Resources == nil || input.Runtime.Resources.Node == nil || input.Runtime.Resources.Node.Enabled == nil {
		return false
	}
	return *input.Runtime.Resources.Node.Enabled
```

internal/otelcollector/config/metric/agent/config_builder.go
k15r: if it impacts performance then please keep it as is!

NHingerl (approved):

k15r (approved):

mmitoraj (approved):

Issue #1506, opened 2024-10-08
secretRef without key does not fail the config generation


a-thaler:
**Description**
When creating a LogPipeline using secretRefs without specifying a key, the pipeline gets accepted and the "ConfigurationGenerated" status gets true. Later on the log delivery fails as no authorization is provided..

Example pipeline:
```yaml
apiVersion: telemetry.kyma-project.io/v1alpha1
kind: LogPipeline
metadata:
  name: test
spec:
  input:
    application:
      containers:
        exclude:
        - istio-proxy
      keepOriginalBody: true
      namespaces: {}
  output:
    http:
      dedot: true
      host:
        valueFrom:
          secretKeyRef:
            key: ingest-endpoint
            name: cls
            namespace: observability
      password:
        valueFrom:
          secretKeyRef:
            key: ingest-password
            name: cls
            namespace: observability
      tls: {}
      uri: /customindex/kyma
      user:
        valueFrom:
          secretKeyRef:
            name: cls
            namespace: observability
status:
  conditions:
  - lastTransitionTime: "2024-10-08T15:33:37Z"
    message: Fluent Bit agent DaemonSet is ready
    reason: AgentReady
    status: "True"
    type: AgentHealthy
  - lastTransitionTime: "2024-10-08T15:34:47Z"
    message: LogPipeline specification is successfully applied to the configuration
      of Fluent Bit agent
    observedGeneration: 2
    reason: AgentConfigured
    status: "True"
    type: ConfigurationGenerated
  - lastTransitionTime: "2024-10-08T15:36:18Z"
    message: 'Backend is not reachable or rejecting logs. All logs are dropped. See
      troubleshooting: https://kyma-project.io/#/telemetry-manager/user/02-logs?id=no-logs-arrive-at-the-backend'
    observedGeneration: 2
    reason: AllTelemetryDataDropped
    status: "False"
    type: TelemetryFlowHealthy
  unsupportedMode: false
```

In the status you see that the configuration is accepted by the telemetry-manager.
I would expect the validation to fail as a "secretKeyRef" without key is meaningless.

Please also verify that the other pipeline types are working as expected.

I observed another scenario where the referenced secret got deleted after the pipeline was running, now the LogPipeline continues to push to a non-existent endpoint instead of failing with configuration problems. Please validate this scenario as well.

PR #1535, opened 2024-10-17
WIP: Secretref validation

TeodorSAP:
## Description

- Set secret name, namespace, and key fields as required

- Resolves https://github.com/kyma-project/telemetry-manager/issues/1506

## Traceability

### DIFF
--- a/apis/telemetry/v1alpha1/shared_types.go
+++ b/apis/telemetry/v1alpha1/shared_types.go
@@ -33,10 +33,13 @@ func (v *ValueFromSource) IsSecretKeyRef() bool {
 
 type SecretKeyRef struct {
 	// The name of the Secret containing the referenced value
+	// +kubebuilder:validation:Required
 	Name string `json:"name,omitempty"`
 	// The name of the Namespace containing the Secret with the referenced value.
+	// +kubebuilder:validation:Required
 	Namespace string `json:"namespace,omitempty"`
 	// The name of the attribute of the Secret holding the referenced value.
+	// +kubebuilder:validation:Required
 	Key string `json:"key,omitempty"`
 }
 
--- a/apis/telemetry/v1alpha1/secret_refs_test.go
+++ b/apis/telemetry/v1alpha1/secret_refs_test.go
@@ -79,6 +79,42 @@ func TestLogPipeline_GetSecretRefs(t *testing.T) {
 				{Name: "creds", Namespace: "default", Key: "password"},
 			},
 		},
+		{
+			name: "http output secret refs (with missing keys)",
+			given: LogPipeline{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "cls",
+				},
+				Spec: LogPipelineSpec{
+					Output: Output{
+						HTTP: &HTTPOutput{
+							Host: ValueType{
+								ValueFrom: &ValueFromSource{
+									SecretKeyRef: &SecretKeyRef{
+										Name: "creds", Namespace: "default",
+									},
+								},
+							},
+							User: ValueType{
+								ValueFrom: &ValueFromSource{
+									SecretKeyRef: &SecretKeyRef{
+										Name: "creds", Namespace: "default",
+									},
+								},
+							},
+							Password: ValueType{
+								ValueFrom: &ValueFromSource{
+									SecretKeyRef: &SecretKeyRef{
+										Name: "creds", Namespace: "default",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: []SecretKeyRef{},
+		},
 	}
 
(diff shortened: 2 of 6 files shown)


apis/telemetry/v1alpha1/shared_types.go
```diff
@@ -29,16 +29,19 @@ type ValueFromSource struct {
 }
 
 func (v *ValueFromSource) IsSecretKeyRef() bool {
-	return v.SecretKeyRef != nil && v.SecretKeyRef.Name != "" && v.SecretKeyRef.Key != ""
+	return v.SecretKeyRef != nil
 }
```
skhalash, line 33: The func name was previously already confusing, but this change made it even worse 😄 Maybe let's remove it?

skhalash (changes requested):

internal/conditions/conditions.go
```diff
@@ -21,7 +21,7 @@ const (
 	ReasonGatewayNotReady             = "GatewayNotReady"
 	ReasonGatewayReady                = "GatewayReady"
 	ReasonMaxPipelinesExceeded        = "MaxPipelinesExceeded"
-	ReasonReferencedSecretMissing     = "ReferencedSecretMissing"
+	ReasonReferencedSecretInvalid     = "ReferencedSecretInvalid"
```
skhalash, line 24: IMHO, the new name does not reflect the reality? Is the secret invalid or secret reference invalid? The new name suggests the former.

apis/telemetry/v1alpha1/secret_refs.go
```diff
@@ -20,7 +20,7 @@ func (lp *LogPipeline) GetEnvSecretRefs() []SecretKeyRef {
 	var refs []SecretKeyRef
 
 	output := lp.Spec.Output
-	if output.IsHTTPDefined() {
+	if output.HTTP != nil {
```
skhalash, line 23: Why this change?

internal/fluentbit/config/builder/output.go
```diff
@@ -107,7 +107,7 @@ func resolveValue(value telemetryv1alpha1.ValueType, logPipeline string) string
 		return value.Value
 	}
 
-	if value.ValueFrom != nil && value.ValueFrom.IsSecretKeyRef() {
+	if value.ValueFrom != nil && value.ValueFrom.SecretKeyRef != nil {
```
skhalash, line 110: We are accessing Name, Namespace, and Key there. Shouldn't we call IsDefined?

apis/telemetry/v1alpha1/shared_types.go
```diff
@@ -20,24 +20,27 @@ func (v *ValueType) IsDefined() bool {
 		return true
 	}
 
-	return v.ValueFrom != nil && v.ValueFrom.IsSecretKeyRef()
+	return v.ValueFrom != nil &&
+		v.ValueFrom.SecretKeyRef != nil &&
+		v.ValueFrom.SecretKeyRef.Name != "" &&
+		v.ValueFrom.SecretKeyRef.Key != "" &&
+		v.ValueFrom.SecretKeyRef.Namespace != ""
```
skhalash, line 27: How about renaming IsDefined to IsValid? I believe it better conveys the intent

apis/telemetry/v1alpha1/secret_refs.go
TeodorSAP: Because the IsHTTPDefined() also checks that the Host field is defined (i.e. is a valid secretref), so then if you have an invalid Host secretref, you won't get the user and the password SRs either, and no error will be triggered, the fields will just be ignored and not appended to the returned SRs.

internal/conditions/conditions.go
TeodorSAP: I agree, but do you have a better suggestion?
Maybe `SecretReferenceInvalid`?

apis/telemetry/v1alpha1/shared_types.go
skhalash: Yeah, everywhere

internal/conditions/conditions.go
skhalash: It should be a union of `SecretReferenceInvalid` and `SecretKeyMissing`. Maybe `UnableToAccessSecret` or something (please, discuss with @a-thaler)

TeodorSAP: I agree with @a-thaler, especially that this is a corner case that will be prevented by the CRD validation in the future. So it's more like the error message that customers currently in this situation see. Given this, changing the name of one of our "most popular" reasons names may not be the best idea.

skhalash (approved):

grego952 (approved):

PR #1557, opened 2024-10-24
feat: Introduce scrape label for usage in NetworkPolicy

jeffreylimnardy:
## Description

- 

- 

## Traceability

### DIFF
--- a/docs/user/04-metrics.md
+++ b/docs/user/04-metrics.md
@@ -729,6 +729,30 @@ To detect and fix such situations, check the pipeline status and check out [Trou
 
 **Remedy 2**: Define the application protocol in the Service port definition by either prefixing the port name with the protocol, like in `http-metrics` or define the `appProtocol` attribute.
 
+**Cause 3**: A deny all `NetworkPolicy` was created in the cluster and the agent cannot scrape metric from annotated workloads.
+
+**Remedy 3**: Define a `NetworkPolicy` to let the agent scrape your workload using the `telemetry.kyma-project.io/metric-scrape` label. An example `NetworkPolicy` could be:
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-traffic-from-agent
+spec:
+  podSelector: 
+    matchLabels:
+      # <your workload here>
+  ingress:
+  - from:
+    - namespaceSelector:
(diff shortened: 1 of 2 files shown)


docs/user/04-metrics.md
```diff
@@ -764,6 +764,32 @@ To detect and fix such situations, check the pipeline status and check out [Trou
 
 **Remedy 2**: Define the application protocol in the Service port definition by either prefixing the port name with the protocol, like in `http-metrics` or define the `appProtocol` attribute.
 
+**Cause 3**: A deny all `NetworkPolicy` was created in the workload namespace and the agent cannot scrape metric from annotated workloads.
```
NHingerl, line 767:
```suggestion
**Cause 3**: A deny-all `NetworkPolicy` was created in the workload namespace, which prevents that the agent can scrape metrics from annotated workloads.
```
If I understand this correctly :)

docs/user/04-metrics.md
```diff
@@ -764,6 +764,32 @@ To detect and fix such situations, check the pipeline status and check out [Trou
...
+
+**Remedy 3**: Define a `NetworkPolicy` to let the agent scrape your workload using the `telemetry.kyma-project.io/metric-scrape` label.
```
NHingerl, line 769: considering that the cause says that a network policy already exists, why does creating a network policy solve the problem?
Are we suggesting to replace the deny-all policy with one that's more permissive?

```suggestion
**Remedy 3**: Replace the deny-all `NetworkPolicy` with one that allows the agent to scrape your workload using the `telemetry.kyma-project.io/metric-scrape` label.
```

jeffreylimnardy: technically, it's possible to modify the deny-all policy to be more permissive, but a more common practice is to define another network policy on top of the deny-all policy, explicitly allowing communication between selected pods, this leaves the deny-all policy untouched and allows for more additional policies to be added separately, but I'll discuss with Icke tomorrow just to be sure that it's what we're going for.

NHingerl: I see, thank you!

NHingerl (dismissed):

rakesh-garimella (approved):

IwonaLanger (approved):

NHingerl:
updated DITA accordingly

Issue #1541, opened 2024-10-21
Migrate // +kubebuilder:default annotations to a defaulting webhook


skhalash:
**Description**

Currently, we are using 

**Reasons**

**Attachments**

**Release Notes**

```

```

hisarbalik:
- LogPipeline validation webhook moved to the kustomize
- Mutating webhook manifest added for all signal types and API version
- Annotation based defaulting logic moved to the new defaulting webhook for all signal types and API versions

PR #1617, opened 2024-11-19
chore: Migrate validating webhook to kustomize

shorim:
## Description

- Let the validating webhook be created by Kustomize and the operator is responsible only for updating it with the CA bundle
- Remove the deletion of validating webhook when Telemetry CR is being deleted, because the validating webhook should now be cleaned up by the [lifecycle-manager](https://github.com/kyma-project/lifecycle-manager) instead of the Telemetry reconciler
- Since the name of the validating webhook is changed from `validation.webhook.telemetry.kyma-project.io` to `telemetry-validating-webhook.kyma-project.io`, the old validating webhook is deleted in the Telemetry reconciler
- For updating the LogPipeline CRD with conversion webhook, use an `update` request instead of `patch` request for simplicity and making it consistent with updating the validating webhook with CA bundle
- Adjust unit and e2e tests

- #1541 

## Traceability

### DIFF
--- a/internal/resources/otelcollector/rbac.go
+++ b/internal/resources/otelcollector/rbac.go
@@ -15,123 +15,218 @@ type Rbac struct {
 	roleBinding        *rbacv1.RoleBinding
 }
 
-func MakeTraceGatewayRBAC(name types.NamespacedName) Rbac {
-	return Rbac{
-		clusterRole:        makeTraceGatewayClusterRole(name),
-		clusterRoleBinding: makeClusterRoleBinding(name),
-		role:               nil,
-		roleBinding:        nil,
+type RBACOption func(*Rbac, types.NamespacedName)
+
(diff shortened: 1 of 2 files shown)


internal/webhookcert/resources.go
```diff
@@ -112,15 +61,13 @@ func makeConversionWebhookConfig(caBundle []byte, config Config) apiextensionsv1
 	}
 }
 
-func patchConversionWebhookConfig(ctx context.Context, c client.Client, conversion apiextensionsv1.CustomResourceConversion) error {
+func updateLogPipelineWithWebhookConfig(ctx context.Context, c client.Client, conversion apiextensionsv1.CustomResourceConversion) error {
```
skhalash, line 64: Would be nice to see something about conversion in the func name. What do you think?

Issue #1644, opened 2024-11-29
chore: Refactor LogPipeline, MetricPipeline, and TracePipeline Controllers


TeodorSAP:
**Description**

The 3 controllers shall be refactored, such that when creating a reconciler, all the variables/constants describing parameters that will not externally change (e.g. dynamic/base CPU, Memory limits) won't have to be passed anymore.

The configurations of the respective gateways and agents have to be refactored accordingly. Such that this pattern is not present in our code anymore.

**Reasons**

1. Bring these variables closer to where they are used, and make the controller code cleaner.
2. Besides the name, namespace, and `priorityClassName` these attributes are not externally configurable anymore (since the corresponding feature flags have been removed)

**Attachments**

Example for LogPipeline:

https://github.com/kyma-project/telemetry-manager/blob/main/controllers/telemetry/logpipeline_controller.go#L69-L87

https://github.com/kyma-project/telemetry-manager/blob/main/controllers/telemetry/logpipeline_controller.go#L175-L192

https://github.com/kyma-project/telemetry-manager/blob/main/controllers/telemetry/logpipeline_controller.go#L224-L242

**Release Notes**

```

```

PR #1659, opened 2024-12-05
chore: Refactor OTel resources

skhalash:
## Description

- The esources/otelcollector package previously defined a "generic" gateway/agent applier-deleter, which required external configuration. However, as gateways for different signal types (e.g., metrics and traces) diverged, the number of configuration parameters increased, leading to complexity.

This PR encapsulates the logic for handling different concrete gateways and agents, simplifying the usage of the package. Additionally, it lays the groundwork for easier integration of log agent support in the future.

- https://github.com/kyma-project/telemetry-manager/issues/1644

## Traceability


.testcoverage.yml
```diff
@@ -35,7 +35,7 @@ override:
     path: ^internal/reconciler/telemetry$
   - threshold: 74
     path: ^internal/reconciler/tracepipeline$
-  - threshold: 82
+  - threshold: 81
```
skhalash, line 38: Sorry for that

internal/utils/test/marshal.go
```diff
@@ -0,0 +1,85 @@
...
+// SaveAsYAML dumps the list of objects as a YAML file. Used to generate golden files, never should be committed.
+func SaveAsYAML(t *testing.T, scheme *runtime.Scheme, objects []client.Object, path string) {
+	objectsYAML, err := MarshalYAML(scheme, objects)
+	require.NoError(t, err)
+
+	err = os.WriteFile(path, objectsYAML, 0600)
+	require.NoError(t, err)
+
+	t.Fatalf("Golden file %s has been saved, please verify it and remove this line", path)
+}
```
k15r, line 58: .... never should be committed?

internal/resources/otelcollector/gateway_test.go
```diff
@@ -2,613 +2,158 @@ package otelcollector
...
+
+			if tt.saveGoldenFile {
+				testutils.SaveAsYAML(t, scheme, objects, tt.goldenFilePath)
+			}
+
+			bytes, err := testutils.MarshalYAML(scheme, objects)
+			require.NoError(t, err)
+
+			goldenFileBytes, err := os.ReadFile(tt.goldenFilePath)
+			require.NoError(t, err)
```
k15r, line 91: im not a big fan of the comparison of those two ginormous strings. 
If we do not implement comparison of the objects, I assume it would be already a big help to finding issues when we at least split the big string in per object strings. That way we can not only print the issue, but also which object was affected.

internal/utils/test/marshal.go
skhalash: Not the function, the usage of the function never should be comitted

internal/resources/otelcollector/gateway_test.go
```diff
@@ -2,613 +2,158 @@ package otelcollector
...
+
+			require.Equal(t, string(goldenFileBytes), string(bytes))
```
k15r, line 93: Can we compare not two ginormous string containing loads of objects, but make it multiple comparisons of "one string per object" and print the name of the object along with the error.

in the future we will probably replace this by comparison of the objects and not their string representation.

internal/utils/test/marshal.go
skhalash: I drew inspiration from https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/pkg/golden/v0.115.0/pkg/golden/golden.go#L113

internal/resources/otelcollector/gateway_test.go
skhalash: Hmm,. we could, but then we need to slice YAMLs by separators :P

controllers/telemetry/logpipeline_controller.go
```diff
@@ -217,47 +201,17 @@ func configureFluentBitReconciler(client client.Client, config LogPipelineContro
 //nolint:unparam // error is always nil: An error could be returned after implementing the IstioStatusChecker (TODO)
 func configureOtelReconciler(client client.Client, config LogPipelineControllerConfig, _ *prober.LogPipelineProber) (*logpipelineotel.Reconciler, error) {
 	otelConfig := logpipelineotel.Config{
-		LogGatewayName:     otelLogGatewayName,
 		TelemetryNamespace: config.TelemetryNamespace,
 	}
```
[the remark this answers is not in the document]
skhalash: Agree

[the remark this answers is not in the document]
skhalash: Done

internal/reconciler/logpipeline/otel/reconciler.go
```diff
@@ -27,7 +26,6 @@ import (
 const defaultReplicaCount int32 = 2
 
 type Config struct {
-	LogGatewayName     string
 	TelemetryNamespace string
 }
```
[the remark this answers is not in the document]
skhalash: Done

internal/resources/otelcollector/gateway_test.go
k15r: It's a bit more complicated but it would improve the readability of the output a lot. (A follow-up issue is also sufficient, so that we don't forget about it)

skhalash: https://github.com/kyma-project/telemetry-manager/issues/1662

k15r (approved):

PR #1736, opened 2025-01-09
chore: Add FluentBit Load Test records/sec columns and re-add generators limits

TeodorSAP:
## Description

- Add records/sec column to FluentBit load test table
- Fix flog CPU limit problem (re-add limit) + avalanche (for metricgen)

Load Test Run: https://github.com/kyma-project/telemetry-manager/actions/runs/12770440464/job/35597795793

- 

## Traceability


hack/load-tests/run-load-test.sh
```diff
@@ -414,8 +414,8 @@ function get_result_and_cleanup_log_otel() {
 function get_result_and_cleanup_fluentbit() {
   RESULT_TYPE="log"
 
-  QUERY_RECEIVED='query=round(sum(rate(fluentbit_input_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024)'
-  QUERY_EXPORTED='query=round(sum(rate(fluentbit_output_proc_bytes_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024)'
+  QUERY_RECEIVED='query=round(sum(rate(fluentbit_input_records_total{service="telemetry-fluent-bit-metrics", name=~"load-test-.*"}[20m])) / 1024)'
```
k15r, line 417: lets not use kiwi-Requests
either use / 1000 (10^3) requests/sec or just use without division

please also adjust the header of the column and even better create a new table with the metrics (maybe even in addition to the old bytes/sec metric)

TeodorSAP: Is it ok like this?

.env
```diff
@@ -16,7 +16,7 @@ ENV_GORELEASER_VERSION=v1.23.0
 
 ## Default Docker Images
 DEFAULT_FLUENTBIT_EXPORTER_IMAGE="europe-docker.pkg.dev/kyma-project/prod/directory-size-exporter:v20241212-e4adf27f"
-DEFAULT_FLUENTBIT_IMAGE="europe-docker.pkg.dev/kyma-project/prod/external/fluent/fluent-bit:3.2.2"
+DEFAULT_FLUENTBIT_IMAGE="europe-docker.pkg.dev/kyma-project/prod/external/fluent/fluent-bit:3.1.9"
```
a-thaler, line 19: Why you are reverting the fluentbit version? To get new numbers for that version it is not required to change the version on main

external-images.yaml
```diff
@@ -1,3 +1,3 @@
 images:
   - source: "fluent/fluent-bit@sha256:b7a2b9ee1958607da9a16b0efeb24a70a43c5b7e91dbabca4dc02e916aadb914"
-    tag: "3.2.2" # used by the kyma telemetry module
```
a-thaler, line 3: That change will have no influence on the Perf setup at all..

hack/load-tests/log-fluentbit-test-setup.yaml
```diff
@@ -99,6 +99,7 @@ spec:
           name: flog
           resources:
             limits:
+              cpu: 50m
```
a-thaler, line 102: In the metric performance tests the load generators are also missing CPU limits? Where they removed as well? Then please revert them as well.

a-thaler (approved):

PR #1782, opened 2025-01-23
chore: Improved Dockerfile of sample-app

a-thaler:
## Description

In order to support multi-platform images for the module, the Dockerfile need to be flexible in regards to the platform input.

- Improved the sample-app dockerfile to not include sources

- 

## Traceability


docs/user/integration/sample-app/Dockerfile
k15r, line 1: on my machine this leads to:
Golang:1.23.5 being downloaded for arm64 (good)
binary compiled for: amd64 (good)
image annotated for: arm64 (BAD?)

a-thaler: I removed the fixed target arch to use the current arch, so that the arch should always fit

PR #1806, opened 2025-02-05
fix: Delete metric agent resources when no longer needed

### LINKED ISSUE #336: Disabled agent features do not remove the agent
**Description**
I have a MetricPipeline (or LogPipeline) defined like that:

```
apiVersion: telemetry.kyma-project.io/v1alpha1
kind: MetricPipeline
metadata:
  name: icke
spec:
  input:
    application:
      runtime:
        enabled: true
  output:
    ...
```
which results in a agent and gateway deployment and all is working as expected.
When disabling the runtime at a later time:
```
  input:
    application:
      runtime:
        enabled: false
```

the agent will not get deleted and continues collecting the runtime data. Indeed, the gateway is filtering out the data and the actual result is as expected. However, removing the agent will save money. Also, the agent uses the last applied config and will send runtime data, but will not scrape istio metrics for example. That is inconsistent.

**Expected result**

The agent gets deleted and stops collecting runtime data.

**Actual result**
The agent stays (as well as the config) in an unattended mode (not being reconciled anymore) and it continues to collect data

**Implementation Hints**
The OTel collector's resource management is in the `internal/resources/otelcollector` pkg. Right now, it has `ApplyAgentResources` and `ApplyGatewayResources` functions available. To add `DeleteAgentResources`, we need to separate making different kinds of resources (RBAC, Services, Deployment/DaemonSet, etc.) from what we do with them (like creating or updating). This change will make it easier to add the option to delete without repeating the code.

**Steps to reproduce**

**Troubleshooting**

TeodorSAP:
## Description

- Delete metric agent resources when no longer needed

- Resolves #336 

## Traceability


internal/reconciler/metricpipeline/reconciler.go
```diff
@@ -166,16 +166,28 @@ func (r *Reconciler) doReconcile(ctx context.Context, pipeline *telemetryv1alpha
...
-		logf.FromContext(ctx).V(1).Info("cleaning up metric pipeline resources: all metric pipelines are non-reconcilable")
+	var reconcilablePipelinesWithAgentRequired []telemetryv1alpha1.MetricPipeline
 
-		if err = r.gatewayApplierDeleter.DeleteResources(ctx, r.Client, r.istioStatusChecker.IsIstioActive(ctx)); err != nil {
-			return fmt.Errorf("failed to delete gateway resources: %w", err)
+	for _, p := range reconcilablePipelines {
+		if isMetricAgentRequired(&p) {
+			reconcilablePipelinesWithAgentRequired = append(reconcilablePipelinesWithAgentRequired, p)
 		}
+	}
```
k15r, line 175: lets introduce a function here to filter the pipelines

test/e2e/metrics_no_input_test.go
```diff
@@ -0,0 +1,103 @@
...
+			Expect(kitk8s.DeleteObjects(ctx, k8sClient, &metricPipelineWithInput)).Should(Succeed())
+		})
+
+		It("Remaining pipeline should have AgentNotRequired condition", Label(suite.LabelUpgrade), func() {
+			assert.MetricPipelineHasCondition(ctx, k8sClient, pipelineNameNoInput, metav1.Condition{
+				Type:   conditions.TypeAgentHealthy,
+				Status: metav1.ConditionTrue,
+				Reason: conditions.ReasonMetricAgentNotRequired,
+			})
+		})
```
k15r, line 97: shouldn't this pipeline always have there metricagentnotrequired condition?

test/testkit/assert/daemon_set.go
```diff
@@ -21,6 +21,13 @@ func DaemonSetReady(ctx context.Context, k8sClient client.Client, name types.Nam
 	}, periodic.EventuallyTimeout, periodic.DefaultInterval).Should(Succeed())
 }
 
+func DaemonSetNotFound(ctx context.Context, k8sClient client.Client, name types.NamespacedName) {
+	Eventually(func(g Gomega) {
+		_, err := isDaemonSetReady(ctx, k8sClient, name)
+		g.Expect(err).To(HaveOccurred())
```
k15r, line 27: please test for IsNotFound error, not any random error

test/e2e/metrics_no_input_test.go
TeodorSAP: It should, but I would still like to test for it. I can move the it statement further up, if it helps.

k15r: yes, please move it up to the setup section of the test.

k15r (approved):

PR #1855, opened 2025-02-17
docs: proposal for switching gateways to a central agent

a-thaler:
## Description

- Proposal on how to change the architecture to mitigate current drawbacks
- Still missing:
  - PoC: Service with local traffic policy, does it work as expected
    - refusing calls if no endpoint is available on same node
    - does it work with Istio in all combinations?
  - Decision: Multiple or one daemonset, is isolation really essential? Can it scale out vertical or is there some limiting factor besides resources?
  - PoC: Test client behaviour, is a small downtime on updates acceptable as clients are doing retries always

- 

## Traceability


docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- separation of concerns
+  It is always good to have clean responsibilities to keep code better maintainable. Here, the gateways will do enrichment and dispatching, while agents to data collection only
+- decoupling of signal types
+  Processing the signal types in separate processes keep the processing tolerant to failures. An overload of metrics will keep the log delivery stable.
+- tail-based sampling
+  Scenarios like trace sampling requires processing all spans of a trace at one instance, requiring a StatefulSet with sticky trace routing. However, that scenario might not be relevant as this will be done on the backend usually.
+
+The drawback of Gateways are:
+- no uniform push endpoint
+  by design there is a gateway per signal type were the application need to communicate to. With that you need to configure an individual OTLP endpoint per signal type always. Having an uniform URL will require an additional component route the requests.
```
mmitoraj, line 27:
```suggestion

  By design, there is one gateway for each signal type that the application must communicate with. Consequently, you must always configure a separate OTLP endpoint for each signal type. Using a uniform URL requires an additional component to route the requests.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- no localhost support
+  A typical approach is to run an agent per node having a host port available where the node-local applications can push the data to. That communication will be very reliable and fast and requires no encryption. Also, the default setting of the otel-sdk will work at any time, no kyma specific endpoint required at all.
```
mmitoraj, line 29:
```suggestion

  A typical approach is to run an agent per node having a host port available where the node-local applications can push the data. That communication is very reliable and fast and requires no encryption. Also, the default setting of the OTel-SDK works at any time, and no Kyma-specific endpoint is required.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- autoscaling is tricky
+  The gateway should provide autoscaling capabilities which are tricky to realize. It must be detected very fast if a scale out is needed and it situations were the backend is overloaded must be excluded from scale out.
```
mmitoraj, line 31:
```suggestion

  The gateway should provide autoscaling capabilities which are tricky to realize. It must be detected very fast if a scale-out is needed and situations where the backend is overloaded must be excluded from the scale-out.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- every instance caches whole cluster state
```
mmitoraj, line 32:
```suggestion
- every instance caches the whole cluster state
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+  Every instance of a gateway must be able to enrich data of every pod and with that need to cache all pod metadata
```
mmitoraj, line 33:
```suggestion

  Every instance of a gateway must be able to enrich the data of every Pod and with that need to cache all Pod metadata.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- Istio mandatory for load balancing
+  Pushing OTLP data from an application to the gateway running with replicas requires loadbalancing. Unfortunately, a kubernetes service is acting on L4 and while GRPC/HTTP2 is on L/, so an additional loadbalancer is needed. That can be easily solved by using Istio, but it becomes a mandatory requirement for applications. However, there can be applications not using Istio (Statefulets often don't do) and like to provide OTLP data.
```
mmitoraj, line 35:
```suggestion

  Pushing OTLP data from an application to the gateway running with replicas requires load balancing. Unfortunately, a Kubernetes service is acting on L4 while GRPC/HTTP2 is on L/, so an additional load balancer is needed. That can be easily solved by using Istio, which becomes a requirement for applications. However, some applications don't use Istio (StatefulSets often don't) but want to provide OTLP data.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- Istio is mandatory for mTLS
+  App to Gateway communication is across nodes and with that should use mTLS, so Istio needs to be used which sometimes is not possible. Also, Istio causes additional overhead like noise filtering
```
mmitoraj, line 37:
```suggestion

  App to Gateway communication is across nodes and with that should use mTLS, so Istio needs to be used which sometimes is not possible. Also, Istio causes additional overhead like noise filtering.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- cumulativetodelta transformation not suppoerted
```
mmitoraj, line 38:
```suggestion
- cumulativetodelta transformation not supported
```
or "cumulative to delta"?

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+  This transformation is mandatory for Dynatrace support and requires that data from a pod gets processed always by the same instance
```
mmitoraj, line 39:
```suggestion

  This transformation is mandatory for Dynatrace support and requires that data from a Pod gets processed always by the same instance.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- General misconception
+  Gateways are usually running outside the cluster as a first entry point. The cluster should get rid of the data as fast as possible as persistence cannot be provided here (thinks like Kafka would again run outside). Try to compensate for short-lived hickups but otherwise just point back the backpressure to the client and do not try to solve the big picture
```
mmitoraj, line 41:
```suggestion

  Gateways are usually running outside the cluster as a first entry point. The cluster should get rid of the data as fast as possible as persistence cannot be provided here (things like Kafka would again run outside). Try to compensate for short-lived hiccups. Otherwise, point back the backpressure to the client and do not try to solve the big picture.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Proposal
+
+Most of the drawbacks can be solved by running the gateway logic node-local only. Every instance always processes only data of the local node. With that a natural scalability is given and can be extended by vertical scaling capabilities. The "old" gateways will be running in agent mode but still called "OTLP Gateway"
```
mmitoraj, line 45:
```suggestion
Most of the drawbacks can be solved by running the gateway logic node-local only. Every instance always processes only data from the local node. With that, a natural scalability is given and can be extended by vertical scaling capabilities. The "old" gateways will be running in agent mode but still called "OTLP Gateway".
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+
+![arch](./../assets/otlp-gateway-new.drawio.svg)
+
+The left drawbacks of the agent approach are:
```
mmitoraj, line 49:
```suggestion
The remaining drawbacks of the agent approach are:
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- coupling of signal types
+  The decoupling of the signal types will be lost, potentially leading to situations where the agent drops logs caused by metrics overload. However, with the right hardening that should not be a problem
```
mmitoraj, line 51:
```suggestion

  The decoupling of the signal types will be lost, potentially leading to situations where the agent drops logs caused by metrics overload. However, with the right hardening that should not be a problem.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- Downtimes on updates
```
mmitoraj, line 52:
```suggestion
- downtimes on updates
```
let's consistently use the lower case

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+  A rollout of the agent cannot happen in a rolling way, especially of the host port assignment. So there will be a downtime always which should be compoensated by the retry mechanism in the otel-sdk
```
mmitoraj, line 53:
```suggestion

  A rollout of the agent cannot happen in a rolling way, especially of the host port assignment. So there will be a downtime always which should be compensated by the retry mechanism in the OTel-SDK.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- host port conflicts
+  The user might serve already the OTLP ports on every port
```
mmitoraj, line 55:
```suggestion

  The user might serve already the OTLP ports on every port.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+
+Additional benefits not outlined yet
```
mmitoraj, line 57:
```suggestion
Additional benefits not outlined yet:
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- persistent queue everywhere
+  A small persistent buffer could be used in all components based on the node file system
```
mmitoraj, line 59:
```suggestion

  A small persistent buffer could be used in all components based on the node file system.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- No Istio dependency
```
mmitoraj, line 60:
```suggestion
- no Istio dependency
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+  There is no Istio dependency at all anymore, no at the App and not at the components. However, the metric agent should still support scraping of istiofied endpoints
```
mmitoraj, line 61:
```suggestion

  Neither the App nor the components have any Istio dependency anymore. However, the metric agent should still support the scraping of istiofied endpoints.
```
You did not mean " istiofied", did you? Should these be "istiofield" instead? I wasn't sure...

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Migration considerations
+
+Unifying the three gateways into one agent should be possible without attention by the user. The API stays compatible, and there should be no direct exposure of the architecture to the user. Still, two problems need to be addressed:
```
mmitoraj, line 65:
```suggestion
Unifying the three gateways into one agent should be possible without user engagement. The API stays compatible, and the architecture shouldn't be directly exposed to the user. Still, two problems must be addressed:
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- Push endpoint compatibility
+  It is not possible to simply switch to t localhost push endpoint, the URLs are used in running software via env variables. So they must stay compatible. Luckily, this can be achieved easily by adjusting the services to point to the new agents using the [internalTrafficPolicy: local](https://kubernetes.io/docs/reference/networking/virtual-ips/#internal-traffic-policy) on the service
```
mmitoraj, line 67:
```suggestion

  It is not possible to simply switch to the localhost push endpoint. The URLs are used in running software via env variables so they must stay compatible. Luckily, this can be achieved easily by adjusting the services to point to the new agents using the [internalTrafficPolicy: local](https://kubernetes.io/docs/reference/networking/virtual-ips/#internal-traffic-policy) on the service.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+- the host ports might be in use already. The manager need to detect that and adjust the module status accordingly. When doing a migration, the situation must be checked upfront.
```
mmitoraj, line 68:
```suggestion
- The host ports are already in use 

  The manager must detect that and adjust the module status accordingly. When migrating, the situation must be checked upfront.
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Conclusion
+
+The former motivation of the gateway concept turned out to be not that relevant. Switching to an agent approach solves a ton of problems while introducing very soft drawbacks. The transformation should start immediately :)
```
mmitoraj, line 72:
```suggestion
The former motivation of the gateway concept turned out to be no longer relevant. Switching to the agent approach solves many problems while introducing very soft drawbacks. The transformation should start immediately :)
```

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,72 @@
...
+
+Proposed
+
+## Context
+
+The current architecture defines a set of central Gateways with the single purpose of enriching the data and dispatching it to the backend. That approach had good reasons in the past but a situation is reached were the drawbacks seem to overweight benefits.
+
+![arch](./../assets/otlp-gateway-old.drawio.svg)
+
+The benefits of Gateways are:
```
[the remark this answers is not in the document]
a-thaler: It is more about that approach/design

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
a-thaler: The Istiofication of workloads is progressing fast, haven't you realized :) I meant it like that :)

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,74 @@
...
+
+Proposed
+
+## Context
+
+The current architecture defines a set of central Gateways to enrich the data and dispatch it to the backend. That approach had good reasons in the past, but now the drawbacks outweigh the benefits.
+
+![arch](./../assets/otlp-gateway-old.drawio.svg)
+
+The benefits of gateways are:
```
mmitoraj, line 15:
```suggestion
### Gateways Benefits

```
I would create subsections for the benefits and drawbacks so that they stick out from the context.

docs/contributor/arch/018-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,74 @@
...
+
+  It is always good to have clear responsibilities to keep code better maintainable. Therefore, gateways deal with enrichment and dispatching, while agents are responsible for data collection.
+- decoupling of signal types
+
+  Processing the signal types in separate processes keeps the processing tolerant to failures. An overload of metrics keeps the log delivery stable.
+- tail-based sampling
+
+  Scenarios like trace sampling require processing all spans of a trace at one instance, requiring a StatefulSet with sticky trace routing. However, that scenario might not be relevant as this isn't usually done on the backend.
+
+The drawbacks of gateways are:
```
mmitoraj, line 29:
```suggestion
### Gateways Drawbacks

```

mmitoraj (approved):

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+  This transformation is mandatory for Dynatrace support and requires that data from a Pod gets processed always by the same instance.
+
+- General misconception
+
+  Gateways are usually running outside the cluster as a first entry point. The cluster should get rid of the data as fast as possible as persistence cannot be provided here (things like Kafka would again run outside). Try to compensate for short-lived hiccups. Otherwise, point back the backpressure to the client and do not try to solve the big picture.
+
+## Proposal
+
+Most of the drawbacks can be solved by running the gateway logic node-local only. Every instance always processes only data of the local node. With that a natural scalability is given and can be extended by vertical scaling capabilities. The "old" gateways will be running in agent mode but still called "OTLP Gateway". Additionally, there will be a new uniform service `telemetry-otlp` forwarding to the node-local entity using the [internalTrafficPolicy: local](https://kubernetes.io/docs/reference/networking/virtual-ips/#internal-traffic-policy) setting. Existing services can stay compatible using the same approach.
```
NHingerl, line 65:
```suggestion
Most of the drawbacks can be solved by running the gateway logic node-locally only. Every instance always processes only data of the local node. With that, a natural scalability is given and can be extended by vertical scaling capabilities. The "old" gateways will be running in agent mode but still called "OTLP Gateway". Additionally, there will be a new uniform service `telemetry-otlp` forwarding to the node-local entity using the [internalTrafficPolicy: local](https://kubernetes.io/docs/reference/networking/virtual-ips/#internal-traffic-policy) setting. Existing services can stay compatible using the same approach.
```

NHingerl, line 65: 'The "old" gateways will be running in agent mode but still called "OTLP Gateway"' 

Isn't this going to be confusing? I trust you have reasons to keep the old name instead of calling them "OTLP Agent", but I'm concerned about the potential for confusion. 
Are we going to document the components as:

> This is the agent, it's called OTLP Gateway.
> These are the gateways, they are called ...node gateway? ...node-local gateway? They're also based on OTLP btw ;)

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Coupling of Signal Types  
+
+Decoupling of signal types will no longer be possible, which may result in scenarios where logs are dropped due to metric overload. However, with proper hardening, this should not pose a significant issue.  
+
+Let's compare a **shared OTLP agent setup** with an **agent-per-signal setup** in terms of robustness. The OTel Collector offers two key mechanisms to prevent overload:
+
+1. **Memory Limiter**  
+   - Applied globally, no way to restrict it to a specific pipeline or signal type.
+   - Performs periodic checks of memory usage and will begin refusing data and forcing GC to reduce memory consumption when defined limits have been exceeded. It influenced by:
```
NHingerl, line 79:
```suggestion
   - Performs periodic checks of memory usage. When defined limits have been exceeded, it begins refusing data and forcing GC to reduce memory consumption. It is influenced by:
```

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+     - Heap allocated object held by the Collector.  It can be the OTLP exporter sending queues or various caches used by different components in the chain (e.g. Kubernetes Attributes Processor informers cache).
```
NHingerl, line 80:
```suggestion
     - Heap-allocated object held by the OTel Collector.  It can be the OTLP exporter sending queues, or various caches used by different components in the chain (like the Kubernetes Attributes Processor informers cache).  
```

NHingerl, line 80: Not sure what the following terms mean, we didn't introduce them earlier:

- **GC** - is this a collector? Gateway Collector?
- **Heap allocated object** - is this an object that's heap-allocated? Then I believe we need a hyphen because "heap allocated" is used as an adjective.
- **Kubernetes Attributes Processor informers cache** - is this the cache of the/an informer of the Kubernetes Attributes Processor? Then maybe it should be "Kubernetes Attributes Processor informer's cache", but even clearer: "Cache of the Kubernetes Attributes Processor informer". Or is it the informer cache of the Kubernetes Attributes Processor? Then drop the "s": "Kubernetes Attributes Processor informer cache"

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+Decoupling of signal types will no longer be possible, which may result in scenarios where logs are dropped due to metric overload. However, with proper hardening, this should not pose a significant issue.  
+
+Let's compare a **shared OTLP agent setup** with an **agent-per-signal setup** in terms of robustness. The OTel Collector offers two key mechanisms to prevent overload:
+
+1. **Memory Limiter**  
+   - Applied globally, no way to restrict it to a specific pipeline or signal type.
+   - Performs periodic checks of memory usage and will begin refusing data and forcing GC to reduce memory consumption when defined limits have been exceeded. It influenced by:
+     - Heap allocated object held by the Collector.  It can be the OTLP exporter sending queues or various caches used by different components in the chain (e.g. Kubernetes Attributes Processor informers cache).  
+     - Heap allocated objects not being used and can be reclaimed by GC.
```
NHingerl, line 81:
```suggestion
     - Heap-allocated objects that aren't being used, which can be reclaimed by GC.
```
Not sure I got that correctly

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+   - Memory Limiter can not prevent all kinds of OOM issues. It can only prevent issues caused by ingested telemetry data by refusing it. For example, the issue of the Kubernetes Attributes Processor informer cache consuming excessive memory can not be prevented by Memory Limiter.  Basically, it is a poor man's throttling mechanism that was necessary when using the old Batch Processor. [Read more](./017-fault-tolerant-otel-logging-setup.md).
```
NHingerl, line 82:
```suggestion
   - Memory Limiter can not prevent all kinds of OOM issues. It can only prevent issues caused by ingested telemetry data by refusing it. For example, Memory Limiter cannot prevent the issue of the Kubernetes Attributes Processor informer cache consuming excessive memory. Basically, it is a poor man's throttling mechanism that was necessary when using the old Batch Processor. See [ADR-017](./017-fault-tolerant-otel-logging-setup.md).
```

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+2. **Refusing Data if OTLP Exporter Queue is Full**  
+   - The Batch Processor must not be used (the built-in OTLP Exporter batcher does not have this limitation and can still be used).
+   - If the queue is full, new data will be refused.  
+   - Each pipeline type has its own queue, allowing different queue lengths to support varying quality-of-service (QoS) levels for different signal types.  
+
+Introducing a shared collector handling OTLP data for all three telemetry types means that only one memory limiter is configured, instead of three separate ones when they are isolated. However, if OTLP exporter queue refusal is correctly set up, the system should never reach the point where the memory limiter starts rejecting data. across all clusters is not solvable by either the old or the new proposed setup.
```
NHingerl, line 89:
```suggestion
Introducing a shared collector that handles OTLP data for all three telemetry types means that only one memory limiter is configured, instead of three separate ones when they are isolated. However, if OTLP exporter queue refusal is correctly set up, the system should never reach the point where the memory limiter starts rejecting data. [MISSING WORDS] across all clusters is not solvable by either the old or the new proposed setup.
```

In front of "across", some words seem to be missing. What is it that isn't solvable?
Maybe something like "Data rejection across all..."

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Downtime During Updates  
+
+Unlike deployments, the agent cannot be rolled out in a rolling manner, meaning updates will always result in some downtime. However, this should be mitigated by the retry mechanism in the OpenTelemetry SDK (according to our [tests](../../contributor/pocs/otelcol-downtime/otelcol-downtime.md) it's not yet the case).
```
NHingerl, line 93:
```suggestion
Unlike deployments, the agent cannot be rolled out in a rolling manner, meaning updates will always result in some downtime. However, this should be mitigated by the retry mechanism in the OpenTelemetry SDK (according to our [tests](../../contributor/pocs/otelcol-downtime/otelcol-downtime.md), it's not yet the case).
```

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+A small persistent buffer can be introduced in all components using the node file system, enhancing resilience and reliability.  
+
+#### No Istio Dependency  
+
+Neither the application nor its components will depend on Istio anymore. However, the metric agent should still support scraping Istio-instrumented endpoints.  
+
+## Conclusion
+
+The former motivation of the gateway concept turned out to be no longer relevant. Switching to the agent approach solves many problems while introducing very soft drawbacks. The transformation should start immediately :)
```
NHingerl, line 107: Exciting times!

NHingerl (commented):
nice!

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+  Pushing OTLP data from an application to the gateway running with replicas requires load balancing. Unfortunately, a Kubernetes service is acting on L4 while GRPC/HTTP2 is on L7, so an additional load balancer is needed. That can be easily solved by using Istio, which becomes a requirement for applications. However, some applications don't use Istio (StatefulSets often don't) but want to provide OTLP data.
+
+- Istio is mandatory for mTLS
+
+  App to Gateway communication is across nodes and with that should use mTLS, so Istio needs to be used which sometimes is not possible. Also, Istio causes additional overhead like noise filtering and collecting observability data about the service mesh should not rely on the service mesh.
+
+- Cumulativetodelta transformation not supported
+
+  This transformation is mandatory for Dynatrace support and requires that data from a Pod gets processed always by the same instance.
```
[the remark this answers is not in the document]
skhalash: Yes, it's this section in particular https://docs.dynatrace.com/docs/ingest-from/opentelemetry/collector/configuration#delta-metrics

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
skhalash: Sorry, it should have been " the Kubernetes Attribute Processor memory usage issue" or something (explained above, the one with the informers cache)

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
skhalash: You got it right

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
a-thaler: It will be hard to rename it as we use the term in our API. A pipeline has a agent and a gateway running.. Maybe we can come up with a completely different term, let's see. I think we don't have to decide now on the name

NHingerl (dismissed):

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+![arch](./../assets/otlp-gateway-new.drawio.svg)
+
+### Remaining Drawbacks of the Agent Approach  
+
+#### Coupling of Signal Types  
+
+Decoupling of signal types will no longer be possible, which may result in scenarios where logs are dropped due to metric overload. However, with proper hardening, this should not pose a significant issue.  
+
+Let's compare a **shared OTLP agent setup** with an **agent-per-signal setup** in terms of robustness. The OTel Collector offers two key mechanisms to prevent overload:
```
TeodorSAP, line 75: (optional) You might consider moving this sentence elsewhere, deleting it, or rephrasing it, since a proper comparison of the two approaches is not explicitly provided, just the explanation of some mechanisms and a paragraph that explains how introducing the new approach might change these mechanisms' configurations and behaviours.

TeodorSAP (changes requested):

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+  Pushing OTLP data from an application to the gateway running with replicas requires load balancing. Unfortunately, a Kubernetes service is acting on L4 while GRPC/HTTP2 is on L7, so an additional load balancer is needed. That can be easily solved by using Istio, which becomes a requirement for applications. However, some applications don't use Istio (StatefulSets often don't) but want to provide OTLP data.
+
+- Istio is mandatory for mTLS
+
+  App-to-gateway communication is across nodes and with that, it should use mTLS; so Istio must be used - which sometimes is impossible. Also, Istio causes additional overhead like noise filtering. Furthermore, collecting observability data about the service mesh should not rely on the service mesh.
+
+- Cumulative to Delta transformation not supported
+
+  Cumulativetodelta transformation is mandatory for Dynatrace support and requires that data from a Pod is always processed by the same instance.
```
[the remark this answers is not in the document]
a-thaler:
```suggestion
  The usage of the `cumulativetodelta` processor is mandatory for Dynatrace support and requires that data from a Pod is always processed by the same instance.
```

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
a-thaler:
```suggestion
To mitigate the coupling problem, an agent per signal type could be used instead of one shared agent. Let's compare a **shared OTLP agent setup** with an **agent-per-signal setup** in terms of robustness. The OTel Collector offers two key mechanisms to prevent overload:
```

docs/contributor/arch/019-switch-from-gateways-to-a-central-agent.md
```diff
@@ -0,0 +1,107 @@
...
+
+- General misconception
+
+  Gateways usually run outside the cluster as a first entry point. The cluster should get rid of the data as fast as possible, because persistence cannot be provided here (things like Kafka would run outside again). Try to compensate for short-lived hiccups. Otherwise, point back the backpressure to the client and do not try to solve the big picture.
```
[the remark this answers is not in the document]
a-thaler:
```suggestion
  The wish of having a dispatching solution inside a cluster, with a configurable persistent buffer and well-defined guarantees is not realistic to achieve with simple kubernetes primitives. A queueing system like Kafka will be required and usually will run as a managed solution outside of the cluster. The elements inside the cluster should focus on processing node-local data as fast as possible and delivering the data to the next service which can provide guarantees.
  Here, we should focus on the agent scenario only by enriching data node-local, leveraging the natural scaling by node instances. In backpressure scenarios, it should not be tried to buffer in a persistent way for a longer time, but instead propagate back the backpressure to the application so that it can react to it.
```

What about that?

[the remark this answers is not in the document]
NHingerl: Icke's suggestion, language-reviewed:

```suggestion
  Using simple Kubernetes primitives, it's unrealistic to create a dispatching solution inside a cluster, with a configurable persistent buffer and well-defined guarantees. A queueing system like Kafka will be required and usually will run as a managed solution outside of the cluster. The elements inside the cluster should focus on processing node-local data as fast as possible and delivering the data to the next service, which can provide guarantees.
  Here, we should focus only on the agent scenario: By enriching data node-locally and using the natural scaling done by node instances. In backpressure scenarios, it should not be tried to buffer in a persistent way for a longer time, but instead, propagate the backpressure back to the application so that it can react to it.
```

NHingerl (approved):

skhalash (approved):

PR #1874, opened 2025-02-25
chore: e2e tests refactoring

TeodorSAP:
## Description

- Split e2e tests into separate sub-folders/packages/suites
- Create and delete the Telemetry CR directly within the GHAs
- Adapt e2e tests accordingly

- 

## Traceability


.github/workflows/pr-integration.yml
```diff
@@ -49,15 +49,24 @@ jobs:
...
+          - filter: "metrics && set_c"
+            path: test/e2e/metrics
+          - filter: traces
+            path: test/e2e/traces
+          - filter: telemetry
+            path: test/e2e/misc
+          - filter: max-pipeline
+            path: test/e2e
+          - filter: misc
+            path: test/e2e/misc
```
k15r, line 69: why do we still need the labels if the sorting is now based on folders?

TeodorSAP: Because there are still e2e tests that use one label but span across multiple subfolders, e.g. `max-pipeline`. And I thought we'll still want this functionality in the future, so there's a bit of flexibility to the subfolders.

TeodorSAP: + There're also scenarios like with upgrade tests, were we just run everything labeled as upgrade, in the whole test/e2e folder recursively.

TeodorSAP: Can you provide me with a clear example of how you would organize them? Since you still have to address certain "edge cases" like:
- `max-pipelines`: that needs to cover more pipeline-types in the same suite
- `logs`: that are technically still logs (covering the same telemetry-type) although now separated in two subfolders (i.e. with fluentbit scheduled for deletion)
- `telemetry`: which covers a small suite of tests, that would fit within the misc sub-folder

So having them perfectly "orthogonal" is not that straight-forward, and maybe not desirable, in certain cases.

TeodorSAP: I gave up on the `path` attribute and am now looking recursively in the whole e2e directory every time (i.e. using labels exclusively, as before); since there are some even "more exceptional" cases where you have, for example, `LabelTraces` cases that need to be executed as part of a `misc` test (see https://github.com/kyma-project/telemetry-manager/blob/main/test/e2e/manager_test.go#L116). So, I have found this approach more permissive in that sense.

.github/workflows/pr-integration.yml
```diff
@@ -207,7 +210,7 @@ jobs:
       - name: Run tests without Istio
         if: ${{ matrix.scenario == 'healthy' }}
         run: |
-          bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="self-mon-${{ matrix.signal-type }}-${{ matrix.scenario }} && experimental" test/e2e
+          bin/ginkgo run ${{ runner.debug && '-v' || '' }} -r --tags e2e --label-filter="self-mon-${{ matrix.signal-type }}-${{ matrix.scenario }} && experimental" test/e2e
```
k15r, line 213:
```suggestion
          bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="self-mon-${{ matrix.signal-type }}-${{ matrix.scenario }} && experimental" -r test/e2e
```

.github/workflows/pr-integration.yml
```diff
@@ -162,7 +164,7 @@ jobs:
       - name: Run tests without Istio
         if: ${{ matrix.scenario == 'healthy' }}
         run: |
-          bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="self-mon-${{ matrix.signal-type }}-${{ matrix.scenario }} && !experimental" test/e2e
+          bin/ginkgo run ${{ runner.debug && '-v' || '' }} -r --tags e2e --label-filter="self-mon-${{ matrix.signal-type }}-${{ matrix.scenario }} && !experimental" test/e2e
```
k15r, line 167:
```suggestion
          bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="self-mon-${{ matrix.signal-type }}-${{ matrix.scenario }} && !experimental" -r test/e2e 
```

.github/workflows/pr-integration.yml
```diff
@@ -187,6 +189,7 @@ jobs:
       fail-fast: false
       matrix:
         signal-type:
+          - logs
```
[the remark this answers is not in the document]
TeodorSAP: What's with this one? I'm confused.

.github/workflows/pr-integration.yml
```diff
@@ -99,7 +100,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Run tests
-        run: bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="${{ matrix.ginkgo-labels }} && experimental" test/e2e
+        run: bin/ginkgo run ${{ runner.debug && '-v' || '' }} -r --tags e2e --label-filter="${{ matrix.ginkgo-labels }} && experimental" test/e2e
```
k15r, line 103:
```suggestion
        run: bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="${{ matrix.ginkgo-labels }} && experimental" -r test/e2e
```

test/e2e/fluentbit/basic_v1alpha1_test.go
k15r, line 1: why do we have a folder called fluentbit and one called logs?
Lets aim for consistency and either have one called logs with subfolders or labels to select the technology
or name them according to the technology used (fluentbit-logs, Otel-logs)

test/e2e/logs/self_monitor_healthy_test.go
```diff
@@ -18,13 +18,13 @@ import (
 	"github.com/kyma-project/telemetry-manager/test/testkit/mocks/backend"
 	"github.com/kyma-project/telemetry-manager/test/testkit/mocks/loggen"
 	"github.com/kyma-project/telemetry-manager/test/testkit/periodic"
-	"github.com/kyma-project/telemetry-manager/test/testkit/suite"
+	. "github.com/kyma-project/telemetry-manager/test/testkit/suite"
 )
 
-var _ = Describe(suite.ID(), Label(suite.LabelExperimental, suite.LabelSelfMonitoringLogsHealthy), Ordered, func() {
+var _ = Describe(ID(), Label(LabelExperimental, LabelSelfMonitoringLogsHealthy), Ordered, func() {
```
k15r, line 24: all other tests in this folder have a rather redundant label called LabelLogs. why do we not have it here? This test is clearly related to logs!

test/e2e/logs/enrichments_extract_labels_test.go
```diff
@@ -20,10 +20,10 @@ import (
 	"github.com/kyma-project/telemetry-manager/test/testkit/mocks/backend"
 	"github.com/kyma-project/telemetry-manager/test/testkit/mocks/telemetrygen"
 	"github.com/kyma-project/telemetry-manager/test/testkit/periodic"
-	"github.com/kyma-project/telemetry-manager/test/testkit/suite"
+	. "github.com/kyma-project/telemetry-manager/test/testkit/suite"
 )
 
-var _ = Describe(suite.ID(), Label(suite.LabelLogs, suite.LabelExperimental), Ordered, func() {
+var _ = Describe(ID(), Label(LabelLogs, LabelExperimental), Ordered, func() {
```
k15r, line 26: lets introduce a label to distinguish tests where a workload pushes its logs and tests where logs are tailed

.github/workflows/pr-integration.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Run tests
-        run: bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="${{ matrix.ginkgo-labels }} && !experimental" test/e2e
+        run: bin/ginkgo run ${{ runner.debug && '-v' || '' }} -r --tags e2e --label-filter="${{ matrix.ginkgo-labels }} && !experimental" test/e2e
```
k15r, line 73:
```suggestion
        run: bin/ginkgo run ${{ runner.debug && '-v' || '' }} --tags e2e --label-filter="${{ matrix.ginkgo-labels }} && !experimental" -r test/e2e
```

test/e2e/fluentbit/basic_v1alpha1_test.go
```diff
@@ -14,25 +14,25 @@ import (
 	kitkyma "github.com/kyma-project/telemetry-manager/test/testkit/kyma"
 	"github.com/kyma-project/telemetry-manager/test/testkit/mocks/backend"
 	"github.com/kyma-project/telemetry-manager/test/testkit/mocks/loggen"
-	"github.com/kyma-project/telemetry-manager/test/testkit/suite"
+	. "github.com/kyma-project/telemetry-manager/test/testkit/suite"
```
k15r, line 17: imho dot imports should be avoided. They make it harder to understand the code. Please consider changing this back

test/e2e/fluentbit/basic_v1alpha1_test.go
TeodorSAP: I thought it would be easier to remove and move the tests when we decide on discontinuing FluentBit, but I can make subfolders, for sure.

test/e2e/logs/enrichments_extract_labels_test.go
TeodorSAP: Should these also be executed as separate suites, like with metrics, or?

test/e2e/fluentbit/basic_v1alpha1_test.go
TeodorSAP: I thought `suite` package variables/functions are used by default in all e2e tests, so it makes it more readable. But if you disagree, I can revert this change, no problem.

test/e2e/logs/self_monitor_healthy_test.go
TeodorSAP: I have not implemented this test, but my guess is that it shouldn't run as part of the Logs suite, but rather just as part of the SelfMon suite/s. I will double-check it though.

test/e2e/fluentbit/basic_v1alpha1_test.go
k15r: imho we should not have a suboptimal layout just because we might remove something in the future.
please either rename the folders or go with the subfolder approach

test/e2e/logs/enrichments_extract_labels_test.go
k15r: yes, that would be great

test/e2e/fluentbit/basic_v1alpha1_test.go
k15r: please revert

test/e2e/logs/self_monitor_healthy_test.go
TeodorSAP: Since these tests are testing the self-mon logic and not the pipelines themselves, I've moved them all to the misc folder. We might even consider creating a separate subfolder just for them. Let me know what you think.

P.S.: This is also the explanation for having a separate label, and running individually from the normal signal-type tests (i.e. they even have a dedicated GHA that executes them individually). My understanding is that this has been done since they need to run in isolation.

k15r (approved):

PR #2117, opened 2025-05-14
fix: move body only if it is json

a-thaler:
## Description

The log body was moved to log.original always, even if it was not parsable

- In case log body is not JSON parsable, skip the whole operator chain as it is based on a parsed body
- Added more advanced test cases
- Removed unneeded operator outputs

- 

## Traceability

### DIFF
--- a/internal/otelcollector/config/log/agent/receivers.go
+++ b/internal/otelcollector/config/log/agent/receivers.go
@@ -196,6 +196,7 @@ func makeDropAttributeLogTag() Operator {
 func makeBodyRouter() Operator {
 	regexPattern := `^{.*}$`
 
+        // If body is not a JSON document, then skip all operators as they are all based on a parsed record and go to noop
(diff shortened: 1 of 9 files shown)


internal/otelcollector/config/log/agent/receivers.go
```diff
@@ -190,16 +193,29 @@ func makeDropAttributeLogTag() Operator {
 	}
 }
 
-// parse body as json and move it to attributes
-func makeJSONParser() Operator {
+func makeBodyRouter() Operator {
 	regexPattern := `^{.*}$`
 
+	return Operator{
```
rakesh-garimella, line 199:
```suggestion
        // If json parsing fails parsing log record, then log record skips all the json parser operators and goes to noop
	return Operator{
```

rakesh-garimella (approved):

Issue #1604, opened 2024-11-12
Streamline MaxPipeline check for LogPipeline


a-thaler:
**Description**
The max pipeline check for LogPipelines is executed as part of the validating webhook. For TracePipeline and MetricPipeline it is executed as a reconciler activity with proper status on the CR.

Goal: Use the same approach for all pipelines and remove unneeded code

Criteria:
- If a pipeline gets added which exceeds the max pipeline threshold, the pipeline never gets active
- The approach is streamlined across the pipeline types
-

PR #2118, opened 2025-05-14
feat: uniform handling of maxpipeline setting

k15r:
## Description

- use pipeline lock for log pipelines
- *IMPORTANT*: ` resourcelock.ErrMaxPipelinesExceeded` will not lead to a reconcile error. As this just leads to more resource consumption due to more reconcile attempts

- https://github.com/kyma-project/telemetry-manager/issues/1604

## Traceability


internal/webhookcert/server_cert_storage_test.go
```diff
@@ -1,24 +1,18 @@
...
 	"github.com/stretchr/testify/require"
 )
 
 func TestStorage(t *testing.T) {
-	certDir := "testdata"
-	require.NoError(t, os.Mkdir(certDir, 0750))
-	sut := serverCertStorageImpl{certDir: certDir}
+	certDir := t.TempDir()
 
-	defer func() {
```
TeodorSAP, line 15: Why are we giving up on the os checks (i.e. directory creation and removal)?

internal/tools/generate
TeodorSAP, line 1: Is this supposed to be here? What is it doing? I don't see it on the main branch.

internal/reconciler/tracepipeline/reconciler_test.go
```diff
@@ -384,7 +384,7 @@ func TestReconcile(t *testing.T) {
 			pipelineValidatorWithStubs,
 			errToMsg)
 		_, err := sut.Reconcile(t.Context(), ctrl.Request{NamespacedName: types.NamespacedName{Name: pipeline.Name}})
-		require.Error(t, err)
+		require.NoError(t, err)
```
TeodorSAP, line 387: Why are we suddenly expecting no error here anymore? Shouldn't the `Reconcile` method return `ErrMaxPipelinesExceeded`, or am I missing something?

TeodorSAP: I noticed the "If a pipeline gets added which exceeds the max pipeline threshold, the pipeline never gets active" requirement. Is this the intended behavior here? Is this the reason for this change and the one from `reconciler.go`, so that we just straight-up skip reconciliation instead of returning `MaxPipelineExceeded`? If yes, can we maybe also assert that reconciliation is skipped, in this test? Maybe that it does not reach a certain execution point?

TeodorSAP: In this UT we just check for the status conditions, would it be possible to maybe have an E2E test where we also check that certain resources are not deployed in this context (for example, that the OTEL ConfigMap does not include the configuration of the 6th pipeline)? Is that already covered?

internal/reconciler/logpipeline/utils/utils.go
TeodorSAP, line 1: Why was an empty utils package created here?

internal/reconciler/telemetry/log_components_checker_test.go
TeodorSAP, line 1: What's the reason behind replacing this test case here? Don't we need both?

internal/reconciler/logpipeline/reconciler_test.go
```diff
@@ -136,6 +138,86 @@ func TestRegisterAndCallRegisteredReconciler(t *testing.T) {
...
+	rec := New(fakeClient, overridesHandler)
+
+	res, err := rec.Reconcile(t.Context(), ctrl.Request{
+		NamespacedName: types.NamespacedName{Name: unsupportedPipeline.Name},
+	})
+	require.ErrorIs(t, err, ErrUnsupportedOutputType)
+	require.Equal(t, ctrl.Result{}, res)
+}
+
+func TestReconcile_LoadingOverridesFailes(t *testing.T) {
```
TeodorSAP, line 202:
```suggestion
func TestReconcile_LoadingOverridesFails(t *testing.T) {
```

internal/utils/logpipeline/logpipeline.go
```diff
@@ -72,3 +77,28 @@ func ContainsCustomPlugin(lp *telemetryv1alpha1.LogPipeline) bool {
 
 	return IsCustomDefined(&lp.Spec.Output)
 }
+
+func GetPipelinesForType(ctx context.Context, client client.Client, mode Mode) ([]telemetryv1alpha1.LogPipeline, error) {
```
TeodorSAP, line 81: What was the need of extracting these 2 functions, since they are only used in the reconciler itself, or?

internal/reconciler/logpipeline/fluentbit/reconciler_test.go
```diff
@@ -761,15 +1004,32 @@ func TestReconcile(t *testing.T) {
 		flowHealthProberStub.On("Probe", mock.Anything, pipeline.Name).Return(prober.LogPipelineProbeResult{}, nil)
 
 		serverErr := errors.New("failed to get secret: server error")
+
+		pipelineLockStub := &mocks.PipelineLock{}
+		pipelineLockStub.On("TryAcquireLock", mock.Anything, mock.Anything).Return(nil)
```
TeodorSAP, line 1009: Are all these tests happy-cases, or are we just not testing any fail-path of `TryAcquireLock` and `IsLockHolder` in the `reconciler_test.go`?

TeodorSAP (changes requested):

test/e2e/logs/suite_test.go
TeodorSAP, line 1: Again you can maybe put this under a shared subpackage, but we are already handling it in our refactoring, please don't invest too much time into it, since this structure and e2e tests in general will change in the near future. See: https://github.com/kyma-project/telemetry-manager/pull/2119

Also, this ADR describes the new structure of E2E tests: https://github.com/kyma-project/telemetry-manager/blob/main/docs/contributor/arch/020-redesign-e2e-log-test-structure.md#miscellaneous

test/e2e/logs/multi_pipeline_max_pipeline_test.go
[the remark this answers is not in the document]
TeodorSAP: If you want to have this test as a single generic one here, that is fine, but then please also remove the one from `fluentbit`

TeodorSAP (changes requested):

test/e2e/logs/suite_test.go
TeodorSAP: Moreover, your changes might cause unwanted conflict and delays on our end, since we explicitly delete tests from these packages that you renamed/moved. Please consider revertin these changes.

test/e2e/logs/multi_pipeline_max_pipeline_test.go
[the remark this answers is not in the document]
k15r: we want all 3

internal/tools/generate
k15r: gone

test/testkit/suite/suite.go
```diff
@@ -100,6 +100,7 @@ const (
 	// Test suites labels
 	LabelLogsOtel      = "logs-otel"
 	LabelLogsFluentBit = "logs-fluentbit"
+	LabelLogs          = "logs"
```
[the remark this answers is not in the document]
k15r: it is needed as some test are testing the reconciler -> sub reconciler setup used for logs

internal/reconciler/telemetry/log_components_checker_test.go
k15r: no reason ... somehow it was modified without me noticing

test/e2e/logs/suite_test.go
TeodorSAP: We've discussed this, potential conflicts will be minimal, we will take care of migrating these new e2e tests on our branch.

TeodorSAP (approved):

PR #2323, opened 2025-07-28
docs: add instructions to scrape metrics from system namespaces

rakesh-garimella:
## Description

- add instructions to scrape metrics from system namespaces

- 

## Traceability


a-thaler:
Is it really just a troubleshooting case? Please verify the attribute description and make it more explicit like at https://kyma-project.io/#/telemetry-manager/user/resources/05-metricpipeline

docs/user/04-metrics.md
```diff
@@ -860,3 +860,32 @@ spec:
 **Cause**: Gateway cannot receive metrics at the given rate.
 
 **Solution**: Manually scale out the gateway by increasing the number of replicas for the metric gateway. See [Module Configuration and Status](https://kyma-project.io/#/telemetry-manager/user/01-manager?id=module-configuration).
+
+### Scrape metrics from system namespaces
+
+**Symptom**: Metrics from system namespaces are not scraped
```
NHingerl, line 866:
```suggestion
**Symptom**: Metrics from system namespaces are not scraped.
```

docs/user/04-metrics.md
```diff
@@ -860,3 +860,32 @@ spec:
 **Cause**: Gateway cannot receive metrics at the given rate.
 
 **Solution**: Manually scale out the gateway by increasing the number of replicas for the metric gateway. See [Module Configuration and Status](https://kyma-project.io/#/telemetry-manager/user/01-manager?id=module-configuration).
+
+### Scrape metrics from system namespaces
```
NHingerl, line 864: The pattern is that the title states the symptom, not the cause.
Consider using similar phrasing to the other troubleshooting sections.
Also, consider putting this one closer to the other "No metrics.." "Not all metrics..." "Only Istio metrics..." sections; I'm not sure it makes sense at the bottom.

```suggestion
### No Metrics From System Namespaces Arrive at the Backend
```

docs/user/04-metrics.md
NHingerl: Is this really the symptom? Is this what the user can observe? 
I would have thought that they can see in the backend that no metrics from system namespaces arrive (similar to the other troubleshooting sections). 
But can they also see what is scraped or not scraped?

docs/user/04-metrics.md
```diff
@@ -860,3 +860,32 @@ spec:
...
+
+**Symptom**: Metrics from system namespaces are not scraped
+
+**Cause**: By default metrics from system namespaces are not scraped.
```
NHingerl, line 868:
```suggestion
**Cause**: This is the default behaviour if you have defined a namespace selector. For details, see [9. Add Filters](#_9-add-filters).
```
Did I understand this correctly? I assume that's because of the filter settings, right?
> The default settings depend on the input:
> If no namespace selector is defined for the prometheus or runtime input, then metrics from system namespaces are excluded by default.
> However, if the namespace selector is not defined for the istio and otlp input, then metrics from system namespaces are included by default.

Also, pls check if the suggested link works.

docs/user/04-metrics.md
```diff
@@ -860,3 +860,32 @@ spec:
...
+
+**Solution**: To scrape metrics from system namespaces, add `namespaces {}` to the respective input section. eg:
```
NHingerl, line 870:
```suggestion
**Solution**: To scrape metrics from system namespaces, add `namespaces {}` to the respective input section; for example:
```

Issue #2123, opened 2025-05-15
Use FIPS compliant images


a-thaler:
**Description**
For using the module in restricted markets a typical requirement is to be compliant with FIPS regulations. With go 1.24.0, FIPS support is [built-in](https://go.dev/doc/security/fips140) into the language and no dynamic linking of ssl libraries is needed.

Goal: All images used by the telemetry module (except of fluentbit) should be running in FIPS mode

Criterias:
- All images are using a FIPS compliant mode (FluentBit will be excluded)
- All images are indicating at startup if they are running actively in that mode (for easy verification)
- Extend documentation to talk about the new compliance
- Assure that all standard backends are still integratable (all integration guides)

Tasks:
- Preparation
  - Build up knowledge about the fips feature in golang (see also https://github.com/kyma-project/lifecycle-manager/issues/2599)
  - Try out the mode on all images and collect problems
  - Try out to connect the otel-collector to backends and collect problems
  - Clarify how to prove that the 3party components of the otel-collector are not doing encryption outside of the golang-sdk (and contiue doing that)
- Implementation
  - [x] Build the otel-collector with FIPS mode and activate it dynamically (https://github.com/kyma-project/telemetry-manager/pull/2603)
  - [x] Build and configure the telemetry-manager with FIPS mode, indicate at startup if the FIPS mode is active (https://github.com/kyma-project/telemetry-manager/pull/2603)
  - [x] Run the tests suites in FIPS mode (https://github.com/kyma-project/telemetry-manager/pull/2631)
  - [x] The self-monitor solution observed FIPS problem in perf tests, we need to figure out the root cause and see if our approach is sustainable long-term
  - [x] Disable http/custom outputs in FIPS mode (https://github.com/kyma-project/telemetry-manager/pull/2691)
  - [x] At startup of the manager, the `KYMA_FIPS_MODE_ENABLED` env variable is read and then
    - the otel-collector will enable the FIPS mode
    - fluentbit is not usable
  - [ ] Build and configure the self-monitor with FIPS mode (https://github.com/kyma-project/telemetry-manager/pull/2603) did not work out and we need to investigate a different solution
    - Decision: we will use a commercial image of prometheus
    - The image will be available at europe-docker.pkg.dev/kyma-project/restricted-prod/sap.com/prometheus-fips:latest
    - The manager should have two env variables, one regular self-monitor image definition and one for the fips image
    - Dependent on the fips more, the manager will take one of the two images
[...]

PR #2603, opened 2025-10-15
feat: enable FIPS mode

shorim:
## Description

- Enable FIPS mode when building the telemetry-manager binary and during its runtime
- Enable FIPS mode when building the self-monitor binary and during its runtime
- If the flag `enable-fips-mode` set to true, then enable FIPS mode for the OTel collectors. Otherwise, disable the FIPS mode for the OTel collectors.
- For the existing metric `telemetry_build_info`, add a label called `fips_mode_enabled` which will be `"true"` if the FIPS mode is enabled and it will be `"false"` otherwise
-  On the startup of the telemetry manager, a log is generated indicating if the FIPS mode is enabled: 
```
{"level":"INFO","timestamp":"2025-10-15T12:36:16Z","logger":"setup","caller":"telemetry-manager-workspace/main.go:278","message":"Starting Telemetry Manager","Build info:":{"fips_mode_enabled":"true","git_commit":"unknown","git_tag":"1.6.0-1461-g0e9d1bf8","git_tree_state":"modified","go_version":"go1.25.3"}}
```

- #2123 

## Traceability

### DIFF
--- a/main.go
+++ b/main.go
@@ -90,7 +90,7 @@ var (
 	enableV1Beta1LogPipelines bool
 	highPriorityClassName     string
 	normalPriorityClassName   string
-	restrictedMode            bool
+	enableFIPSMode            bool
 )
 
 const (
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -79,18 +79,18 @@ var (
...
 	otelCollectorImage     string
 	selfMonitorImage       string
 	alpineImage            string
+
+	// Operator flags
+	certDir                   string
+	enableV1Beta1LogPipelines bool
+	highPriorityClassName     string
+	normalPriorityClassName   string
+	isRestrictedMarket        bool
```
jeffreylimnardy: can we make this a bit more generic? since it's kyma is open source, and restricted market is not a familiar term maybe for everyone?

shorim: I'm open for ideas. What other name do you have in mind?

PR #2687, opened 2025-11-13
chore: introduce global config

skhalash:
## Description

- Introduce a global config, carrying runtime-defined properties, which are cross-cutting by nature to avoid passing and copying the same values over and over again (the way we did with the target/manager namespace)
- Move some constants closer to the usage instead of passing them in the configs
- Fix cached upgrade test problem by appending -count=1 flag, which invalidates the cache

- 

## Traceability

### DIFF
--- a/.github/workflows/pr-upgrade.yml
+++ b/.github/workflows/pr-upgrade.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Run tests on latest tag
         shell: bash
         run: |
-          go test -v -timeout=10m -run="^(TestLogsFluentBitUpgrade|TestLogsUpgrade|TestMetricsUpgrade|TestTracesUpgrade)/before_upgrade$" ./test/e2e/upgrade/...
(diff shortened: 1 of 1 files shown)


internal/config/global.go
```diff
@@ -0,0 +1,66 @@
...
+
+// WithNamespace sets both TargetNamespace, DefaultTelemetryNamespace and ManagerNamespace to the given value.
+// TODO: Split into separate options.
+func WithNamespace(namespace string) Option {
+	return func(g *Global) {
+		g.namespace = namespace
+	}
+}
+
+func WithOperateInFIPSMode(enable bool) Option {
```
k15r: lets remove the enable parameter. it is by default false and the option just enables it

internal/selfmonitor/config/testdata/config.yaml
```diff
@@ -3,7 +3,8 @@ global:
     evaluation_interval: 30s
 alerting:
     alertmanagers:
-        - static_configs:
+        - scheme: https
```
k15r: why do we set the default to `https` just to directly after it only use hosts with `http`?

k15r (approved):

PR #2739, opened 2025-12-03
chore: return generic error for user when validating webhook fails unexpectedly

jeffreylimnardy:
## Description

- `ottl.NewTransformSpecValidator` and `ottl.NewFilterSpecValidator` are not dependent on user input and thus errors returned by these functions must be generic to not confuse user but detailed error messages should be provided as logs for maintainers to identify issues

- 

## Traceability

### DIFF
--- a/.github/workflows/pr-upgrade.yml
+++ b/.github/workflows/pr-upgrade.yml
@@ -43,19 +43,26 @@ jobs:
       - name: Setup Golang
         uses: "./.github/template/setup-golang"
 
-      - name: Checkout latest tag
+      - name: Checkout latest release
         id: latest-tag
         shell: bash
         run: |
-          git fetch --tags
-          LATEST_TAG=$(git tag --sort=-creatordate | sed -n 1p)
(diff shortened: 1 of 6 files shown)


webhook/utils/transformfilter.go
```diff
@@ -24,7 +27,8 @@ func ValidateFilterTransform(signalType ottl.SignalType, filterSpec []telemetryv
 
 	transformValidator, err := ottl.NewTransformSpecValidator(signalType)
 	if err != nil {
-		return fmt.Errorf("failed to instantiate TransformSpecValidator %w", err)
+		logf.Log.V(1).Error(err, "Failed to instantiate TransformSpec validator")
+		return fmt.Errorf("failed to create pipeline")
```
skhalash: 1. Your errors message is a static string - it is better to use `errors.New`.
2. Let's create a private variable since it is used in 2 places.

skhalash (changes requested):

Issue #2729, opened 2025-12-01
Migrate from v1alpha1 to v1beta1


rakesh-garimella:
### Description
- Change the reconciler to use v1Beta1
- update makefile `make manifests`
- Update Busola to use new versions only
- Change to rollout `v1beta1` version
  ```yaml
   versions:
   - name: v1alpha1
     served: true
     storage: false
     deprecated: true
     deprecatedWarning: "some-text"
   - name: v1beta1
     served: true
     storage: true
  ```
- Do a manual upgrade test
- Prepare documentation PR incl. migration guide

> Careful dont turn `served: false` as deletion and creation of v1alpha1 resources would fail

jeffreylimnardy:
post-rollout will be handled in a separate issue, see #767

PR #2781, opened 2026-01-07
feat: telemetry  v1beta1 migration

jeffreylimnardy:
## Description

- Change all the reconciliation components, unit tests, and e2e tests to use v1beta1 instead of v1alpha1
- Remove v1beta1 feature flag
- Switched storage version to beta

- #2729 

## Traceability


internal/fluentbit/config/builder/output.go
```diff
@@ -77,35 +77,35 @@ func generateHTTPOutput(httpOutput *telemetryv1alpha1.FluentBitHTTPOutput, fsBuf
 	}
 
 	tlsEnabled := "on"
-	if httpOutput.TLS.Disabled {
+	if httpOutput.TLSConfig.Insecure {
```
skhalash: I just realized that the corresponding field in the OTLPOutput (of the same type) is called `TLS`, not `TLSConfig`, and is also a pointer. It was also called `TLS` in v1alpha1. 

Do we really need to rename this field?

internal/otelcollector/config/logagent/filelog_receiver.go
```diff
@@ -95,12 +95,17 @@ func createExcludePath(application *telemetryv1alpha1.LogPipelineApplicationInpu
...
+		}
+
+		if runtime.Containers != nil {
+			excludeContainers = append(excludeContainers, runtime.Containers.Exclude...)
+		}
 	}
 
-	if application == nil || (!application.Namespaces.System && len(application.Namespaces.Include) == 0 && len(application.Namespaces.Exclude) == 0) {
+	if runtime == nil || runtime.Namespaces == nil || (len(runtime.Namespaces.Include) == 0 && len(runtime.Namespaces.Exclude) == 0) {
 		excludeNamespaces = namespaces.System()
```
skhalash: I see that you removed some unit tests. Do we still have a test case that covers this scenario?

skhalash (dismissed):

apis/telemetry/v1beta1/logpipeline_types.go
NHingerl: line 93: Should be "an OTLP source," not "a..."
line 142: Should be "a FluentBitHTTP-based...", not "an".

skhalash (approved):

PR #2868, opened 2026-01-13
docs: adjust documentation for v1beta1

jeffreylimnardy:
## Description

- Rename all apiVersion fields to use beta instead of alpha
- Rename application to runtime in all examples of LogPipelines
- Rename spec.input.otlp.disabled to enabled, flipping the logic where necessary
- Rewrite system namespaces filtering docs in Filter Logs page

- 

## Traceability


a-thaler (dismissed):

a-thaler:
Lets wait with the merge shortly before release to not have it published already

docs/user/collecting-logs/runtime-input.md
```diff
@@ -1,6 +1,6 @@
 # Configure Application Logs
```
NHingerl: Shouldn't we change the title of this doc, too? Readers might be confused if we sometimes call that runtime logs and sometimes application logs. Or is that by purpose?

jeffreylimnardy: I'm actually not sure about that part, I thought it's technically still application logs that users are collecting, but we just now rename the field to runtime, but it's true that users might be confused if we have that inconsistency

NHingerl (approved):

Issue #2165, opened 2025-05-27
Align Service Name enrichment


a-thaler:
## Description

The otel-collector recently introduced proper service attributes (`service.namespace`, `service.name`, `service.version`, `service.instance.id`) enrichment as part of the [`k8sattributes` processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md) (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39335/files).

The conventions and enrichment fallback chains that are implemented by OTel are documented here: https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes

Currently, we are performing the enrichment in a custom way by using our custom [`servicenameenrichment` processor](https://github.com/kyma-project/opentelemetry-collector-components/tree/main/processor/serviceenrichmentprocessor). Therefore, we should aim to use the upstream feature and eliminate the custom logic.

A **first challenge** here is that the Istio trace spans are enriched with the service name attribute by Istio itself, following custom Istio logic.

A **second challenge** is represented by the fact that this change will be breaking for our current users, as currently enriched telemetry data does not follow the same OTel convention (fallback chain logic) for setting the service attributes.

> [!NOTE]
> See ADR for more details: https://github.com/kyma-project/telemetry-manager/pull/2870

## Goal

Use the upstream enrichment feature (that is OTel's `k8sattributes` processor) for enriching service attributes (`service.namespace`, `service.name`, `service.version`, `service.instance.id`) and ensure consistency across all exported telemetry data.

## Criterias
- Enrichment logic is now taking care of the additional attributes: `service.namespace`, `service.version`, and `service.instance.id`
- No custom enrichment logic is present
- Istio enrichment is overwritten according to OTel's standards
- Breaking changes are handled properly

## Tasklist
- [ ] Phase 1: Introducing the New Enrichment Logic (https://github.com/kyma-project/telemetry-manager/issues/2890)
- [ ] Phase 2:
  - [ ] Configure the default Telemetry CR, so that newly created resources have the annotation set to `k8sattributes` by default
  - [ ] Add a warning banner to Kyma Dashboard, that warns users about this future deprecation
  - [ ] Document deprecation process
  - [ ] Implement a custom metric that reflects this feature adoption rate
  - [ ] Update Plutono dashboards to track this new metric
- [ ] Phase 3: Deprecation with Backward Compatibility
[...]

PR #2870, opened 2026-01-14
docs: service attributes enrichment using consistent OTel approach proposal ADR

TeodorSAP:
## Description

- Proposal ADR: "Service Attributes Enrichment using Consistent OTel Approach"

- https://github.com/kyma-project/telemetry-manager/issues/2165

## Traceability


docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+
+A first challenge here is that the Istio trace spans are enriched with the service name attribute by Istio itself, following custom Istio logic. In our case, this logic is configured in the [MeshConfig of the Kyma Istio module](https://github.com/kyma-project/istio/blob/6295e154b3992cf42c44d40eed3c2ec488f990bf/internal/istiooperator/istio-operator.yaml#L237), in which the `TracingServiceName` field is set to `CANONICAL_NAME_ONLY` (https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig-TracingServiceName). This, in turn, enriches the `service.name` attribute of Istio-generated trace spans with the canonical name for a workload, which results in the following fallback chain enrichment logic (in order):
+- `service.istio.io/canonical-name` label (https://github.com/istio/istio/blob/master/pkg/model/proxy.go#L492)
+- `app.kubernetes.io/name` label
+- `app` label
+- `"istio-proxy"`
+
+A second challenge is represented by the fact that this change will be breaking for our current users, as currently enriched telemetry data does not follow the same OTel convention (fallback logic) for setting the service attributes.
```
NHingerl:
```suggestion
As a second challenge, this would be a breaking change for our current users, because currently enriched telemetry data does not follow the same OTel convention (fallback logic) for setting the service attributes.
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+### First Challenge: Overwriting Istio Trace Spans
+As we actively enrich Istio data, we should overwrite the Istio enrichment (as a documented feature) to make that enrichment consistent across everything.
+
+In order to correctly identify Istio-generated trace spans, there are mainly two options:
+1. Adding a custom attribute to the span (from Istio's side)
+2. Using Istio-specific attributes that are already set on the span
+
+#### Option 1: Setting A Custom Attribute
+
+Although this approach would be more consistent with our current way of handling Istio access logs, as of today there are no configuration options on Istio's side that would allow setting a custom attribute for OTel tracing.
```
NHingerl:
```suggestion
Although this approach is more consistent with our current way of handling Istio access logs, Istio currently offers no configuration options to set a custom attribute for OTel tracing.
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+
+In this initial phase, the annotation is introduced and automatically set to `k8sattributes` for newly created Telemetry resources. Existing Telemetry resources have the annotation unset, defaulting to the legacy `servicenameenrichment` processor to ensure no breaking changes.
+
+**Default Behavior:**
+- **New Telemetry resources**: Annotation set to `k8sattributes` (uses new processor)
+- **Existing Telemetry resources (annotation unset)**: Use `servicenameenrichment` processor (preserves existing behavior)
+- **Warning condition**: Added to the Telemetry CR status for existing resources, recommending migration to the new processor
+
+> [!NOTE]
+> The Kyma lifecycle-manager applies the default Telemetry CR only once during module enablement. Subsequent module upgrades do not overwrite existing Telemetry resources, preserving custom configurations and annotations. An edge case exists when users manually delete the Telemetry CR, causing the lifecycle-manager to reapply the default configuration on the next reconciliation. This behavior will be addressed in future improvements.
```
NHingerl:
```suggestion
> The Kyma Lifecycle Manager (KLM) applies the default Telemetry CR only once during module enablement. Subsequent module upgrades do not overwrite existing Telemetry resources, preserving custom configurations and annotations. As an edge case, if users manually delete the Telemetry CR, KLM reapplies the default configuration on the next reconciliation. This behavior will be addressed in future improvements.
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+
+**User Action (Optional):**
+Users can proactively adopt the new processor by setting the annotation. The warning condition in the Telemetry CR status will be cleared once the annotation is explicitly set:
```
NHingerl:
```suggestion
If users want to adopt the new processor proactively, they can set the annotation manually, which also removes the warning condition in the Telemetry CR status:
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Phase 2: Deprecation with Backward Compatibility
+
+In this phase, the default behavior changes: resources with unset annotations now use the `k8sattributes` processor. However, users can still explicitly opt into the legacy `servicenameenrichment` processor if additional migration time is needed.
+
+**Default Behavior (Annotation Unset):**
+- **All Telemetry resources**: Use `k8sattributes` processor by default
+- **Enhanced warnings**: Telemetry CRs with the legacy processor explicitly set will show stronger deprecation warnings
+
+**User Action (Optional):**
```
NHingerl: Again, I'd switch the paragraphs for User Action and Status Condition

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+Users requiring additional time for migration can temporarily revert to the legacy processor using the annotation:
```
NHingerl:
```suggestion
Users that need more time for migration can temporarily revert to the legacy processor with the following annotation:
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+  name: sample
+  annotations:
+    # Temporarily opt back into legacy processor
+    kyma-project.io/telemetry-service-enrichment: servicenameenrichment
+spec:
+  # ...
+```
+
+**Status Condition:**
+If annotation is set to `servicenameenrichment`, then the Telemetry CR will include a warning condition notifying users that the `servicenameenrichment` processor is deprecated and will be removed in the future.
```
NHingerl:
```suggestion
If the annotation is set to `servicenameenrichment`, then the Telemetry CR includes a warning condition notifying users that the `servicenameenrichment` processor is deprecated and will be removed in the future.
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+We are already using the `component` attribute in order to identify Istio trace spans in the `istionoisefilter` processor:
+```go
+// component must be "proxy" to be considered an Istio proxy span.
+isIstioProxy := attrs.component == "proxy"
+```
+> Source: https://github.com/TeodorSAP/opentelemetry-collector-components/blob/test/empty-service-name/processor/istionoisefilter/internal/rules/span.go#L12
+
+Once identified, these trace spans can be easily dropped, by using an OTel transform processor before the `k8sattributes` processor, which will then enrich them properly. See example below:
```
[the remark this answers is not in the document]
TeodorSAP: Yes, exactly. The idea is to first drop the enrichment generated automatically by Istio (i.e. that sets the `service.name` attribute in its spans), and then use the new enrichment logic from the OTel `k8sprocessor`, which enriches `service.name`, `service.namespace`, etc. according to its own logic (that is, a different fallback chain used for naming these attributes)

This way we make sure that all telemetry data is enriched uniformly in accordance with OTel's enrichment logic.

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,204 @@
...
+date: 2026-01-12
+---
+
+# 30: Service Attributes Enrichment using Consistent OTel Approach
+
+## Context
+
+The OTel Collector recently introduced proper service attributes (`service.namespace`, `service.name`, `service.version`, `service.instance.id`) enrichment as part of the `k8sattributes` processor (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39335/files).
+
+For the conventions and enrichment fallback chains see [OpenTelemetry: Service attributes](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes).
```
NHingerl:
```suggestion
For the conventions and enrichment fallback chains, see [OpenTelemetry: Service attributes](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes).
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,204 @@
...
+title: Service Attributes Enrichment using Consistent OTel Approach
+status: Proposed
+date: 2026-01-12
+---
+
+# 30: Service Attributes Enrichment using Consistent OTel Approach
+
+## Context
+
+The OTel Collector recently introduced proper service attributes (`service.namespace`, `service.name`, `service.version`, `service.instance.id`) enrichment as part of the `k8sattributes` processor (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39335/files).
```
NHingerl: Sorry for touching this again, but the long bracket between "attributes" and "enrichment" bugged me too much :D

```suggestion
With [PR 39335](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39335/files), the Otel Collector `k8sattributes` processor introduced enrichment with service attributes (`service.namespace`, `service.name`, `service.version`, `service.instance.id`).
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,204 @@
...
+
+For the conventions and enrichment fallback chains see [OpenTelemetry: Service attributes](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes).
+
+Currently, we are performing the enrichment in a custom way by using our custom `servicenameenrichment` processor. Therefore, we should aim to use the upstream feature and eliminate the custom logic.
```
NHingerl:
```suggestion
Currently, we use our custom `servicenameenrichment` processor for this enrichment. We should use the upstream feature and eliminate our custom logic. 
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,204 @@
...
+2. `app.kubernetes.io/name` label
+3. `app` label
+4. `"istio-proxy"`
+
+As a second challenge, this would be a breaking change for our current users, because currently enriched telemetry data does not follow the same OTel convention (fallback logic) for setting the service attributes.
+
+## Proposal
+
+### First Challenge: Overwriting Istio Trace Spans
+Since Telemetry Manager actively enriches Istio data, we should overwrite the default Istio service attributes enrichment (as a documented feature) to ensure a uniform enrichment logic across every telemetry data, according to the OTel conventions.
```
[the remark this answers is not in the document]
TeodorSAP: I wouldn't drop the first part. Or add it as an argument in a second sentence.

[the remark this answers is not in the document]
NHingerl: I see! How about this?

```suggestion
Already in our current implementation, Telemetry Manager modifies Istio data. Overwriting the default Istio service attributes according to OTel conventions is a necessary addition to ensure uniform enrichment logic.
We must provide user documentation about the new enrichment logic and the temporary annotation used during the deprecation.
```

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,204 @@
...
+2. `app.kubernetes.io/name` label
+3. `app` label
+4. `"istio-proxy"`
+
+As a second challenge, this is a breaking change for our users, because the new OTel convention for setting service attributes differs from our current implementation.
+
+## Proposal
+
+### First Challenge: Overwriting Istio Trace Spans
+Since Telemetry Manager actively enriches Istio data, we should overwrite the default Istio service attributes enrichment (as a documented feature) to ensure a uniform enrichment logic across every telemetry data, according to the OTel conventions.
```
a-thaler: The manager is not enriching anything, it is the otel-collector. Also, why this is the reason that we should do a overwrite of the data?

The best solution is that Istio follows the convention or can be configured in a way that it follows the convention. As this is not the case, first of all you should create a ticket for Istio to request that, as then there would be nothing to do. As Istio does not follow the convention yet (maybe it will do after fixing your ticket request), we need to compensate that by actively modifying the data. The best place is the otel-collector which anyway is already dealing with Istio spans.

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,204 @@
...
+      - delete_key(resource.attributes, "service.name") where span.attributes["component"] == "proxy"
+```
+
+### Second Challenge: Incrementally Introducing This Breaking Change
+
+To minimize disruption to existing users, we roll out the migration from the custom `servicenameenrichment` processor to the standard OTel `k8sattributes` processor in three phases. This approach ensures backward compatibility and gives users time to adapt to the new behavior.
+
+#### Annotation-Based Processor Selection
+
+On the Telemetry CR, we introduce a custom annotation `kyma-project.io/telemetry-service-enrichment` to control which service enrichment processor is applied across all telemetry types and pipelines (traces, metrics, and logs). This annotation accepts the following values:
```
a-thaler: We already have quite some labels in place like "telemetry.kyma-project.io/metric-export". Please keep the naming consistent

docs/contributor/arch/030-service-attributes-enrichment.md
```diff
@@ -0,0 +1,204 @@
...
+**Default Behavior:**
+- **New Telemetry resources**: Annotation set to `k8sattributes` (uses new processor)
+- **Existing Telemetry resources (annotation unset)**: Use `servicenameenrichment` processor (preserves existing behavior)
+- **Warning condition**: Added to the Telemetry CR status for existing resources, recommending migration to the new processor
+
+> [!NOTE]
+> The Kyma Lifecycle Manager (KLM) applies the default Telemetry CR only once during module enablement. Subsequent module upgrades do not overwrite existing Telemetry resources, preserving custom configurations and annotations. As an edge case, if users manually delete the Telemetry CR, KLM reapplies the default configuration on the next reconciliation. This behavior will be addressed in future improvements.
+
+**Status Condition:**
+The Telemetry CR will include a warning condition similar to `CertAboutToExpire` indicating the deprecation status and recommending adoption of the new processor.
```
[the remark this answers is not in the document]
TeodorSAP: It might be too early for a warning, but how can we make the users aware of this recommended opt-in in phase 1 otherwise? In phase 2, this behavior will already be changed without them knowing, which might lead to confusion.

docs/contributor/arch/030-service-attributes-enrichment.md
TeodorSAP: Lmk what you think about the new rephrasing.

Issue #2890, opened 2026-01-19
Align Service Name enrichment: Activatable new enrichment logic


a-thaler:
**Description**

First phase of https://github.com/kyma-project/telemetry-manager/issues/2165

  - [ ] Implement custom transform processor for dropping istio trace spans' service enrichment attributes
  - [ ] Configure `k8sattributes` processor to also enrich service attributes
  - [ ] Implement custom Telemetry CR annotation for switching between the legacy enrichment logic (`servicenameenrichment`) and the new one (istio enrichment drop + `k8sattributes`)

PR #2906, opened 2026-01-22
feat: implement configurable enrichment strategy behavior through custom annotation

TeodorSAP:
## Description

- Implement configurable enrichment strategy behavior through custom annotation
- Add tip in user docs
- Add golden file tests
- Refactor some existing e2e tests
- Implement e2e tests for the new otel enrichment strategy
- Refactor reconcilers' code to avoid duplicated functions
- Simplify `update-golden-files` Make target

- Resolves https://github.com/kyma-project/telemetry-manager/issues/2890
- Phase 1 of https://github.com/kyma-project/telemetry-manager/issues/2165

## Traceability

### DIFF
--- a/external-images.yaml
+++ b/external-images.yaml
@@ -1,10 +1,10 @@
 images:
   - source: "fluent/fluent-bit@sha256:7d727245767ae632eb296c2ff4d206bf2e205b5f244c1f37b8fdd61f9fb33985"
     tag: "4.2.2" # used by the kyma telemetry module
-  - source: "rancher/k3s@sha256:f9f125ef9c662a231a98c507afdd3ba9a94d5f02631f946d1d34000ef67f7263"
-    tag: "v1.32.8-k3s1" # used by k3d - current k8s version
-  - source: "rancher/k3s@sha256:f9f125ef9c662a231a98c507afdd3ba9a94d5f02631f946d1d34000ef67f7263"
-    tag: "v1.33.4-k3s1" # used by k3d - future k3s version
+  - source: "rancher/k3s@sha256:bfced830bf3a9b4624e0e569264c7bfebc0700bc36d20a7238faa8646825e36b"
+    tag: "v1.33.7-k3s1" # used by k3d - current k8s version
+  - source: "rancher/k3s@sha256:71abd3a56f57884c62732e0e0d87606052cb5f8555b7db7e8e33c04570b8175c"
+    tag: "v1.34.3-k3s1" # used by k3d - future k8s version
   - source: "library/registry@sha256:3725021071ec9383eb3d87ddbdff9ed602439b3f7c958c9c2fb941049ea6531d"
     tag: "3" # used by k3d
   - source: "library/alpine@sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62"


docs/user/filter-and-process/automatic-data-enrichment.md
```diff
@@ -19,6 +19,11 @@ The gateway determines the service name based on the following hierarchy of labe
 4. Pod name
 5. If none of the above is available, the value is `unknown_service`
 
+> [!WARNING]
+> The behavior described above for enriching unspecified or unknown service names will be deprecated; enable the new OTel-consistent service enrichment strategy by setting the `telemetry.kyma-project.io/service-enrichment` annotation in the Telemetry CR to `otel`, which enriches `service.namespace`, `service.name`, `service.version`, and `service.instance.id` according to the [OTel semantic convention](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes).
```
NHingerl:
```suggestion
> The Telemetry module also supports enrichment with service attributes matching OTel conventions (see [OTel: Service Attributes](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes), which enriches `service.namespace`, `service.name`, `service.version`, and `service.instance.id`.
If you'd like to use that, manually set the `telemetry.kyma-project.io/service-enrichment` annotation in the Telemetry CR to `otel`. If you want to return to the previous method, set the annotation back to `kyma-legacy`.
```

docs/user/filter-and-process/automatic-data-enrichment.md
```diff
@@ -19,6 +19,11 @@ The gateway determines the service name based on the following hierarchy of labe
...
+> 
+> By default, the legacy strategy is currently used, but this behavior will change in future releases, with the new OTel strategy becoming the default. If you, for any reason, want to continue using the legacy strategy, you can do that for a limited amount of time during the deprecation process by explicitly setting the annotation to `kyma-legacy`.
```
NHingerl: No. All of this is in the future, let's keep it out of our docs as long as it isn't the case yet. As long as it's not deprecated, let's avoid an informal pre-deprecation somewhere in the docs.

```suggestion

```

TeodorSAP: Moreover, I would like to make it clear that setting it to `kyma-legacy` is not advisable at all, but it's an option if they want to stick with the legacy strategy for longer. This functionality is already implemented in this PR. The only thing that this PR does not yet do is change the defaults, which will be done in the near future.

docs/user/filter-and-process/automatic-data-enrichment.md
TeodorSAP: I agree we could make it sound less severe, but I would still highlight that it's not "just another option", but rather a configuration that they're advised/encouraged to do, since it will become the future default at some point in the near future.

test/e2e/traces/service_enrichment_test.go
```diff
@@ -0,0 +1,181 @@
...
+		telemetrygen.WithServiceName(unknownServicePattern))
+	podSpecWithCustomServiceAttributes := telemetrygen.PodSpec(telemetrygen.SignalTypeTraces,
+		telemetrygen.WithServiceName(customServiceName),
+		telemetrygen.WithServiceNamespace(customServiceNamespace),
+		telemetrygen.WithServiceVersion(customServiceVersion),
+		telemetrygen.WithServiceInstanceID(customServiceInstanceID),
+	)
+
+	// Enable OTel service enrichment strategy
+	// TODO(TeodorSAP): Remove this block after deprecation period ends and OTel strategy becomes default enrichment strategy
```
jeffreylimnardy: lets reset the telemetry CR after tests are done to prevent potential flakiness
```suggestion
	// TODO(TeodorSAP): Remove this block after deprecation period ends and OTel strategy becomes default enrichment strategy
    kitk8s.PreserveAndScheduleRestoreOfTelemetryResource(t, kitkyma.TelemetryName)
```

docs/user/filter-and-process/automatic-data-enrichment.md
```diff
@@ -24,6 +24,12 @@ The gateway determines the service name based on the following hierarchy of labe
 > 
 > If you'd like to use that, manually set the `telemetry.kyma-project.io/service-enrichment` annotation in the Telemetry CR to `otel`. If you want to return to the previous method, set the annotation back to `kyma-legacy`.
 
+> [!NOTE]
```
NHingerl: Image

Half of the section "Service Name" is now a tip, followed by a note. Does this still serve the purpose of this section? Mind, this is the document about Automatic Data Enrichment, and the section about how the service name is automatically enriched. 

The first tip added the information that users can manually switch the automatic enrichment method. Now we're adding edge cases. This is derailing the scope of this document and won't meet users' needs.

Do you really expect a) users who consider to switching the enrichment method and b) the few users who happen to experience the edge case will look for that information in a topic called "Automatic Data Enrichment"? This used to be an informational topic about what our module does automatically, but it's below the surface it's evolving into something instructional, with a faint hint of migration guide. 

Which user need are we trying to meet here?

NHingerl (approved):

PR #3380, opened 2026-04-13
fix: allow OAuth2 tokenURL to use secret references

a-thaler:
## Summary

The CEL validation expression on the `OAuth2Options.TokenURL` field prevented using `valueFrom` (secret references) for the token URL. The expression used `self.value != ''` to guard the `isURL()` check, which can behave unexpectedly in Kubernetes CEL evaluation when the `value` field is absent due to `omitempty`. This caused the API server to reject pipelines with `tokenURL` defined via a secret reference.

## Changes

- **Fixed CEL expression** in both `v1beta1` and `v1alpha1` API types: changed from `(self.value != '' ) ? (isURL(self.value)) : true` to `has(self.value) ? isURL(self.value) : true`. Using `has()` is the correct way to check field presence in Kubernetes CEL — when `value` is absent (because `valueFrom` is used), `has(self.value)` definitively returns `false`, skipping URL validation.
- **Regenerated CRDs** for both default and experimental Helm charts (all 3 pipeline types × 2 API versions).
- **Added E2E acceptance tests** for all pipeline types (traces, metrics, logs) verifying that OAuth2 configuration with all fields (`clientID`, `clientSecret`, `tokenURL`) provided via secret references is accepted by the API server.

## Test plan

- [x] All unit tests pass (`make test`)
- [x] E2E tests: new `TestAcceptTracePipelineCreation`, `TestAcceptMetricPipelineCreation`, and `TestAcceptLogPipelineCreation` validate that pipelines with OAuth2 secret references are accepted
- [x] Existing `otlp-output-oauth2-invalid-token-url` rejection tests continue to catch invalid direct URL values


test/e2e/logs/misc/reject_creation_test.go
```diff
@@ -16,6 +16,37 @@ import (
 	"github.com/kyma-project/telemetry-manager/test/testkit/suite"
 )
 
+func TestAcceptLogPipelineCreation(t *testing.T) {
```
TeodorSAP: Feels weird having "AcceptCreation" tests in reject_creation_test.go. It would be easier and better to just change the normal "oauth2_test.go" tests to use secrets instead of a direct value.

a-thaler: done as requested

TeodorSAP (approved):

Issue #3384, opened 2026-04-13
Add VPA for OTLP gateway


a-thaler:
**Description**
Similar to https://github.com/kyma-project/telemetry-manager/pull/3114 the otlp gateway should scale vertical on demand.

Criterias:
- VPA gets created/deleted together with the gateway daemonset
- There is a meaningful max value configured
- It gets disabled by annotation and has meaningful memroy limits in place when being disabled
- Align the annotation behaviour across the VPAs
- The ratio of request and limit should be similar to the other VPAs
- Initial request should be as minimal as possible

**Related Items**
- https://github.com/kyma-project/telemetry-manager/issues/424
- https://github.com/kyma-project/telemetry-manager/pull/3114

PR #3387, opened 2026-04-14
feat: add VPA support for OTLP gateway

hisarbalik:
## Description

- VPA enabled by default, gets disabled by annotation `telemetry.kyma-project.io/enable-vpa` value `false`
- Delete VPA CRs when the gateway is being deleted
- Set maxAllowed memory in the VPA to 30% of the allocatable memory of the smallest node
- Set memory request/limit ratio to factor 2
- Add user documentation for VPA

- https://github.com/kyma-project/telemetry-manager/issues/3384

## Traceability


docs/user/pipeline-vpa.md
```diff
@@ -0,0 +1,134 @@
+# Vertical Pod Autoscaler (VPA) for Telemetry Pipelines
```
shorim:
```suggestion
# Vertical Pod Autoscaler (VPA) for Telemetry Components
```

docs/user/manage-pipeline-resources.md
```diff
@@ -0,0 +1,104 @@
...
+## Prerequisites
+
+- Cluster Nodes have sufficient allocatable memory to accommodate resource adjustments.
+- The VPA Custom Resource Definition (CRD) is installed in your cluster. Most managed Kubernetes distributions (for example, GKE, EKS, AKS) provide VPA as an optional feature. For installation instructions, see [Vertical Pod Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler). If the VPA CRD is not available, Telemetry Manager automatically falls back to static resource configuration. Your pipelines continue to work.
+
+## Context
+
+The Telemetry module configures VPA with the following settings:
+
+- Memory limits are set to 2x the memory request.
```
shorim, line 14: Only the initial memory limit of the resource is twice the memory reques, right?
And that is not a configuration of the VPA itself. So, I would remove it from here or what do you think?

docs/user/manage-pipeline-resources.md
```diff
@@ -0,0 +1,104 @@
...
+- Maximum allowed memory is calculated as 30% of the smallest Node's allocatable memory, rounded down to the nearest KiB.
+
+When you upgrade the Telemetry module to a version with VPA support, VPA activates for existing Telemetry pipelines. This triggers a one-time restart of Telemetry components as VPA applies its initial resource recommendations. During the initial adjustment period, you might observe Pod restarts as VPA learns your workload patterns. This is normal behavior.
```
shorim:
```suggestion
When you upgrade the Telemetry module to a version with VPA support, VPA activates for existing Telemetry components. This triggers a one-time restart of Telemetry components as VPA applies its initial resource recommendations. During the initial adjustment period, you might observe Pod restarts as VPA learns your workload patterns. This is normal behavior.
```

docs/user/manage-pipeline-resources.md
```diff
@@ -0,0 +1,104 @@
...
+           Memory:  64Mi
+         Target:
+           Memory:  128Mi
+         Upper Bound:
+           Memory:  256Mi
+   ```
+
+## Result
+
+When VPA is disabled, Telemetry components use static resource configurations, scaling memory requests based on the number of active pipelines using a fixed multiplier.
```
shorim: Is this also true for metric agent and log agent? I think there if VPA is disabled, then they have a memory limit of 1200 Mi regardless of the number of pipelines

docs/user/troubleshooting.md
```diff
@@ -213,3 +213,37 @@ filter:
 ### Solution
 
 Review the syntax of your transform and filter rules and ensure that the names of [OTTL functions](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs/README.md) are spelled correctly (for example, `IsMatch()` instead of `isMatch()`).
+
+## VPA Not Creating Resources
```
shorim:
```suggestion
## VPA Resources Are Not Created
```

docs/user/troubleshooting.md
```diff
@@ -213,3 +213,37 @@ filter:
...
+
+### Solution
+
+If data is not arriving at your backend, see [Not All Data Arrive at the Backend](#not-all-data-arrive-at-the-backend) to check for backend-side issues first.
+
+If data is arriving but components are running out of memory, consider one of these options:
+
+- Add nodes with more memory to increase the maxAllowed calculation.
+- Reduce telemetry volume by applying filters in your pipelines (see [Filter Logs](./filter-and-process/filter-logs.md), [Filter Traces](./filter-and-process/filter-traces.md), [Filter Metrics](./filter-and-process/filter-metrics.md)).
+- Disable VPA, which causes the system to use static resource configuration with a fixed multiplier (see [Manage Automatic Resource Scaling](./manage-pipeline-resources.md)).
```
shorim: same idea here that the "fixed multiplier" is not true for the metric agent and log agent

docs/user/manage-pipeline-resources.md
hisarbalik: No, every time VPA scales the memory, VPA will generate a recommendation for the request and set a limit to 2x the request

docs/user/manage-pipeline-resources.md
hisarbalik: No, agents having static resources configured independent from number of pipelines. I will remove it

docs/user/troubleshooting.md
hisarbalik: Changed

internal/reconciler/otlpgateway/reconciler.go
```diff
@@ -363,9 +395,14 @@ func (r *Reconciler) cleanupLegacyGateways(ctx context.Context) error {
...
 
+	vpaCRDExists, err := r.vpaStatusChecker.VpaCRDExists(ctx, r.Client)
+	if err != nil {
+		return fmt.Errorf("failed to check VPA CRD: %w", err)
+	}
+
 	ns := r.globals.TargetNamespace()
 	for _, gatewayName := range []string{names.TraceGateway, names.MetricGateway, names.LogGateway} {
-		if err := otelcollector.DeleteLegacyGatewayResources(ctx, r.Client, ns, gatewayName, isIstioActive); err != nil {
+		if err := otelcollector.DeleteLegacyGatewayResources(ctx, r.Client, ns, gatewayName, isIstioActive, vpaCRDExists); err != nil {
```
shorim: The legacy gateways don't have a corresponding VPA, so we don't need to attempt deleting VPAs for them

hisarbalik: removed

internal/reconciler/otlpgateway/reconciler_test.go
```diff
@@ -35,6 +36,8 @@ type mocks struct {
 	gatewayApplierDeleter *mockGatewayApplierDeleter
 	configBuilder         *mockOTLPGatewayConfigBuilder
 	istioStatusChecker    *mockIstioStatusChecker
+	vpaStatusChecker      *mockVpaStatusChecker
+	nodeSizeTracker       *mockNodeSizeTracker
```
shorim: vpaStatusChecker and nodeSizeTracker don't need to be mocks. Let's make them stubs similar to what we have done for the metricpipeline https://github.com/kyma-project/telemetry-manager/tree/main/internal/reconciler/metricpipeline/stubs and logpipeline https://github.com/kyma-project/telemetry-manager/tree/main/internal/reconciler/logpipeline/stubs

shorim: And actually the existing gatewayApplierDeleter and configBuilder mocks should be moved to the dedicated mocks folder and the existing istioStatusChecker should be changed to a stub and moved to a stubs folder, but I informed @TeodorSAP about that and he will do it in a separate PR. But can you change in this PR the vpaStatusChecker and nodeSizeTracker into stubs and move them into a stubs folder? @hisarbalik

internal/resources/otelcollector/legacy_gateway_cleanup_test.go
```diff
@@ -0,0 +1,107 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestDeleteLegacyGatewayResources(t *testing.T) {
+	tests := []struct {
+		name            string
+		istioActive     bool
+		vpaCRDExists    bool
```
shorim: we don't need this attribute, since legacy gateways didn't have a VPA

internal/resources/otelcollector/legacy_gateway_cleanup_test.go
```diff
@@ -0,0 +1,107 @@
...
+		gatewayName     string
+		createResources bool
+	}{
+		{
+			name:            "delete with istio and VPA",
```
shorim:
```suggestion
			name:            "delete with istio",
```

internal/resources/otelcollector/legacy_gateway_cleanup_test.go
```diff
@@ -0,0 +1,107 @@
...
+			createResources: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			scheme := runtime.NewScheme()
+			utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+			utilruntime.Must(istiosecurityclientv1.AddToScheme(scheme))
+			utilruntime.Must(autoscalingvpav1.AddToScheme(scheme))
```
shorim: we should not need this now as we are not attempting to delete a VPA

internal/resources/otelcollector/legacy_gateway_cleanup_test.go
```diff
@@ -0,0 +1,107 @@
...
+				require.NoError(t, fakeClient.Create(context.Background(), deployment))
+
+				secret := &corev1.Secret{ObjectMeta: objectMeta}
+				require.NoError(t, fakeClient.Create(context.Background(), secret))
+
+				configMap := &corev1.ConfigMap{ObjectMeta: objectMeta}
+				require.NoError(t, fakeClient.Create(context.Background(), configMap))
+			}
+
+			err := DeleteLegacyGatewayResources(context.Background(), fakeClient, "kyma-system", tt.gatewayName, tt.istioActive, tt.vpaCRDExists)
```
shorim:
```suggestion
			err := DeleteLegacyGatewayResources(context.Background(), fakeClient, "kyma-system", tt.gatewayName, tt.istioActive)
```

internal/resources/otelcollector/legacy_gateway_cleanup_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+			// The function should not return an error (it's idempotent)
+			// But it may return errors if deletion fails, which is acceptable
+			if err != nil {
```
shorim: It should not return an error unless there is a server error. So, let's assert that the error is nil

internal/resources/otelcollector/legacy_gateway_cleanup_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			scheme := runtime.NewScheme()
+			utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+			utilruntime.Must(istiosecurityclientv1.AddToScheme(scheme))
+			utilruntime.Must(autoscalingvpav1.AddToScheme(scheme))
+
+			fakeClient := fake.NewClientBuilder().WithScheme(scheme).Build()
+
+			if tt.createResources {
+				// Create mock legacy resources
```
shorim: The resources are defined twice. First time when creating them and second time when checking that they were deleted. Let's define a list of resources once and re-use it.

shorim: Also, there are many resources missing here: common resources (RBAC resources (ServiceAccount, ClusterRole, ClusterRoleBinding, Role, RoleBinding) and metrics Service), network policies and peerauthentication (if istio is active)

internal/resources/otelcollector/otlp_gateway.go
```diff
@@ -251,6 +263,35 @@ func (o *OTLPGatewayApplierDeleter) DeleteResources(ctx context.Context, c clien
...
 }
 
+// applyVPA creates, updates, or deletes the VPA resource based on configuration.
+func (o *OTLPGatewayApplierDeleter) applyVPA(ctx context.Context, c client.Client, labelerClient client.Client, name types.NamespacedName, opts GatewayApplyOptions) error {
+	if !opts.VpaCRDExists {
+		return nil
+	}
+
+	if opts.VpaEnabled {
+		vpa := makeVPA(name, "DaemonSet", o.baseMemoryRequest, opts.VPAMaxAllowedMemory)
```
shorim: since all of the components having a VPA now are DaemonSets, we can remove this argument and just always set the `Kind` as DaemonSet in the `makeVPA` func

internal/resources/otelcollector/otlp_gateway_test.go
```diff
@@ -39,7 +42,7 @@ func TestOTLPGateway_ApplyResources(t *testing.T) {
 	// Interface for testing both gateway types
 	type gatewayApplierDeleter interface {
 		ApplyResources(ctx context.Context, c client.Client, opts GatewayApplyOptions) error
-		DeleteResources(ctx context.Context, c client.Client, isIstioActive bool) error
+		DeleteResources(ctx context.Context, c client.Client, isIstioActive bool, vpaCRDExists bool) error
 	}
 
 	tests := []struct {
```
shorim, line 47: let's add a test entries here so that we can see how the VPA looks like in golden file tests. Similar to what we have in the agents test https://github.com/kyma-project/telemetry-manager/blob/main/internal/resources/otelcollector/agent_test.go#L68-L81

internal/resources/otelcollector/otlp_gateway_test.go
```diff
@@ -223,3 +230,192 @@ func TestOTLPGateway_Annotations(t *testing.T) {
...
+			})
+			require.NoError(t, err)
+
+			require.Equal(t, tt.expectVPACreate, createdVPA, "VPA create expectation mismatch")
+			require.Equal(t, tt.expectVPADelete, deletedVPA, "VPA delete expectation mismatch")
+		})
+	}
+}
+
+func TestOTLPGateway_ResourceRequirements(t *testing.T) {
```
shorim: I think once we add the golden test file when the VPA is enabled, we won't need this unit test

internal/resources/otelcollector/otlp_gateway_test.go
```diff
@@ -223,3 +230,192 @@ func TestOTLPGateway_Annotations(t *testing.T) {
...
+				memLimit := resources.Limits[corev1.ResourceMemory]
+				expectedLimit := memRequest.DeepCopy()
+				expectedLimit.Add(memRequest)
+				require.Equal(t, expectedLimit.String(), memLimit.String())
+			}
+		})
+	}
+}
+
+func TestOTLPGateway_MakeGatewayMetadata(t *testing.T) {
```
shorim: Do we need this test? Isn't it already covered from the `TestOTLPGateway_ApplyResources`? when the `istioEnabled` is true, we can see the generated metadata in the golden test file

shorim: Also, I see another existing unit test called  `TestOTLPGateway_Annotations`. Isn't that also already covered by the golden test file? If yes, please also remove it

internal/reconciler/otlpgateway/reconciler.go
shorim, line 390: We are not setting the new `VpaActive` field at all, so it will always be false in this case. It should be true when `vpaCRDExists && isVpaEnabled`

internal/otelcollector/config/otlpgateway/config_builder_test.go
```diff
@@ -28,7 +28,50 @@ func TestBuild(t *testing.T) {
 		metricPipelines   []telemetryv1beta1.MetricPipeline
 		serviceEnrichment string
 		moduleVersion     string
+		vpaActive         bool
 	}{
+		{
```
shorim: do we really need 4 different unit tests for checking that the `gomemlimit` is disabled in the `cgroup_runtime` extension when VPA is active? It is the same extension defined once regardless of which pipeleines exist. So, let's have only one unit test for that. Maybe one test having the 3 signal pipelines

internal/resources/otelcollector/otlp_gateway.go
```diff
@@ -387,6 +428,15 @@ func (o *OTLPGatewayApplierDeleter) makeGatewayResourceRequirements(opts Gateway
 		cpuRequest.Add(o.dynamicCPURequest)
 	}
 
+	// When VPA is active, override the memory limit to 2x the memory request so the VPA can scale within a tighter range.
+	// This replaces the calculated memory limit with a value based on the memory request.
+	// For more details, check the ADR: https://github.com/kyma-project/telemetry-manager/blob/main/docs/contributor/arch/032-vertical-pod-autoscaler-VPA-architecture.md
+	if opts.VpaCRDExists && opts.VpaEnabled {
+		vpaMemoryLimit := o.baseMemoryRequest.DeepCopy()
```
shorim, line 436:
```suggestion
        memoryRequest := o.baseMemoryRequest.DeepCopy()
		vpaMemoryLimit := o.baseMemoryRequest.DeepCopy()
```

There are 2 issues here if we don't reset the memoryRequest to the baseMemoryRequest:
1- the memoryRequest which we set the resource requirements will be the multiplied value instead of the base value even when VPA is active
2- the memory limit when vpa is active will be `baseMemoryRequest + multiplied memoryRequest` instead of `2 * baseMemoryRequest`

shorim: let's add a unit test in the `TestOTLPGateway_ApplyResources` where VPA is active and the `ResourceRequirementsMultiplier` is not 0, to test the behavior of this part

internal/otelcollector/config/otlpgateway/config_builder_test.go
hisarbalik: There are 4 config scenarios, gor each signal type and mix all of them

shorim: yeah, but is it really needed to have these 4 scenarios? I think only one is enough. Because when the VPA is active only the `extensions` part is affected which is common for all pipelines

internal/resources/otelcollector/otlp_gateway.go
shorim: ah, but the `otlpGatewayDynamicMemoryRequest` is 0, so even with multiple pipelines we add 0 to the memory request, so will not be affected. However, let's make it more clear by resetting the memory request to the base value

NHingerl (dismissed):

shorim (approved):

NHingerl (approved):

NHingerl:
must add this to Help Portal

PR #3408, opened 2026-04-16
chore: support PR-based self-monitor image in load test workflow

a-thaler:
## Summary
- Add a boolean `selfmonitor_pr_image` checkbox to the PR Load Test workflow to use the self-monitor image built from the PR
- Add explicit `--set manager.container.env.selfMonitorImage` to the `deploy-no-fips` Makefile target so the image can be overridden via the `SELF_MONITOR_IMAGE` env var (defaults to the production image from `.env`)
- Print diagnostic output in case a component is not rolling out in time

### DIFF
--- a/.github/workflows/pr-loadtest.yml
+++ b/.github/workflows/pr-loadtest.yml
@@ -24,6 +24,10 @@ on:
         description: "Duration of the test in seconds"
         required: true
         default: "1200"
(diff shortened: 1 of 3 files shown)


k15r (approved):
fyi: i still dont like our run-load-test.sh

Issue #2922, opened 2026-01-29
Align Service Name enrichment: New clusters will use new logic


a-thaler:
**Description**
Second phase of https://github.com/kyma-project/telemetry-manager/issues/2165

  - [x] Configure the default Telemetry CR, so that newly created resources have the annotation set to `k8sattributes` by default
  - [x] Add a warning banner to Kyma Dashboard in case annotation is not present with new value, that warns users about the future deprecation 
  - [x] Document deprecation process
  - [x] Implement a custom metric that reflects this feature adoption rate
  - [x] Update dashboards to track this new metric

PR #3428, opened 2026-04-22
feat: new clusters will use new strategy for enriching service attributes

shorim:
## Description

- For the default Telemetry CR, set the annotation `telemetry.kyma-project.io/service-enrichment` to `otel`. This Telemetry CR will be applied in new clusters, resulting in the new service attributes enrichment strategy being used in these new clusters
- Add a warning in Kyma Dashboard about the future deprecation in case the annotation `telemetry.kyma-project.io/service-enrichment` is not set to  `otel`
- Document that the legacy enrichment strategy will be deprecated
- Add a custom metric that reflects which service attributes enrichment strategy is used

- #2922 

## Traceability


TeodorSAP:
@NHingerl, will we need a deprecation note for this? Or just in the next phase, when we actually deprecate the legacy annotation?

helm/files/telemetries-busola-config.yaml
```diff
@@ -85,6 +85,13 @@ details: |-
             header: Last transition time
             type: TraceComponentsHealthy
   body:
+    - name: Deprecation of legacy strategy for service name enrichment
+      widget: Panel
+      visibility: $not($lookup(metadata.annotations, "telemetry.kyma-project.io/service-enrichment") = "otel")
+      children:
+      - widget: Alert
+        severity: warning
+        source: "'The legacy strategy for service name enrichment will be deprecated. To avoid disruptions, migrate to the new OTel-based enrichment strategy by adding the annotation `telemetry.kyma-project.io/service-enrichment: otel` to your Telemetry CR. For more details, check https://kyma-project.io/external-content/telemetry-manager/docs/user/filter-and-process/automatic-data-enrichment.html#service-attributes.'"
```
TeodorSAP: Have you checked the Busola UI after the changes?

telemetry-default-cr.yaml
```diff
@@ -5,3 +5,5 @@ apiVersion: operator.kyma-project.io/v1beta1
 kind: Telemetry
 metadata:
   name: default
+  annotations:
+    telemetry.kyma-project.io/service-enrichment: otel
```
TeodorSAP, line 9: I thought we decided to move this to the dist folder, or?

internal/reconciler/telemetry/test_helpers_test.go
```diff
@@ -0,0 +1,94 @@
+package telemetry
```
TeodorSAP, line 1: Why don't we call this file just test_helpers? Since it does not contain any actual tests. Same for logpipeline, metricpipeline, and tracepipeline packages in reconciler.

TeodorSAP (changes requested):

helm/files/telemetries-busola-config.yaml
shorim: yes, let me paste a screenshot here

telemetry-default-cr.yaml
shorim: yeah, I will do that in a separate PR, because it will affect many other related parts when we also move the helm chart under a dist folder

internal/reconciler/telemetry/test_helpers_test.go
shorim: Good question! I followed the naming pattern we already have in the other reconcilers packages. My only concern if we change it to not have the `_test.go` suffix is that we might have some workflows which are triggered or not triggered based on this pattern. I will check this out.

helm/files/telemetries-busola-config.yaml
shorim: Screenshot 2026-04-23 at 16 22 00

docs/user/filter-and-process/automatic-data-enrichment.md
```diff
@@ -26,6 +26,9 @@ The gateway determines the service name based on the following hierarchy of labe
 >
 > However, if you choose to use the OTel enrichment strategy, be aware of [these OTel-specific edge case limitations](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md#configuring-recommended-resource-attributes).
 
+> [!WARNING]
+> The legacy strategy for service name enrichment will be deprecated. To avoid disruptions, migrate to the new OTel-based enrichment strategy by adding the annotation `telemetry.kyma-project.io/service-enrichment: otel` to your Telemetry CR.
```
NHingerl: I'm assuming that it will already be deprecated when we release 1.63.
Working on the deprecation note.

```suggestion
> The legacy strategy for service name enrichment is deprecated. To avoid disruptions, migrate to the new OTel-based enrichment strategy by adding the annotation `telemetry.kyma-project.io/service-enrichment: otel` to your Telemetry CR.
```
Instead of spelling out how to migrate here, can we link to the document that describes the new enrichment strategy?

internal/reconciler/telemetry/test_helpers_test.go
shorim: As expected, the `pr-integration` workflow ignore changes in files with the `_test.go` suffix, because we don't want to run this workflow if we only have changes in unit tests. Note that it has these 2 regular expressions `!**_test.go` and `test/**.go`. So, it will ignore changes only in unit tests, but still get triggered for changes in our e2e tests.

Also, Go compiler doesn't include test files in the generated binary and it does that relying on the test files ending with the "_test.go" suffix. So, we should keep the "_test.go" suffix to avoid including these files in the generated binary

docs/user/filter-and-process/automatic-data-enrichment.md
shorim: @NHingerl It won't be considered deprecated in release 1.63. We will deprecate it in the future with phase 3 here https://github.com/kyma-project/telemetry-manager/issues/2165. Probably actual deprecation will be in release 1.64. We should want to give an early warning and give users more time to switch to the new strategy

NHingerl: "Deprecation Note" is the warning. 
"Deletion Note" is what we use to announce we'll remove the feature in the future.

shorim: As agreed, I changed the warning now to `is deprecated` instead of `will be deprecated`.

shorim: The description of the new enrichment strategy is in the same document

NHingerl (approved):

TeodorSAP (approved):

PR #3475, opened 2026-05-05
chore: requeue with specific interval for when self monitor is not ready

k15r:
## What Changed

All four pipeline reconcilers (FluentBit log, OTel log, metric, and trace) now gate flow health probing on self-monitor Deployment readiness. Before probing flow health, each reconciler checks whether the self-monitor is ready; if it is not deployed, the reconciler sets `FlowHealthy` to `Unknown` without requeueing, and if the Deployment exists but is not yet ready, the reconciler requeues after 30 seconds. This PR also fixes a potential negative TLS certificate requeue duration and corrects early-return logic in the Telemetry reconciler.

## Affected Signal Types

Logs, Metrics, Traces

<details>
<summary>Key Changes</summary>

- **`internal/reconciler/commonstatus`**: Adds the shared `CheckSelfMonitorReadiness` function and the `RequeueDelayOnFlowHealthProbingFailure` constant (30 seconds) used by all pipeline reconcilers to gate flow health probing on self-monitor availability.
- **`controllers/telemetry`**: Wires a `DeploymentProber` as `selfMonitorProber` into all four pipeline reconcilers via the new `WithSelfMonitorProber` option.
- **`internal/reconciler/logpipeline/fluentbit`, `internal/reconciler/logpipeline/otel`, `internal/reconciler/metricpipeline`, `internal/reconciler/tracepipeline` (status)**: Adds a self-monitor readiness check in `evaluateFlowHealthCondition`; if the self-monitor Deployment is not found, the reconciler sets `FlowHealthy` to `Unknown` without requeueing, and if the Deployment exists but is not ready, the reconciler sets `flowHealthProbingFailed = true` to trigger a timed requeue.
- **`internal/reconciler/logpipeline/fluentbit`, `internal/reconciler/logpipeline/otel`, `internal/reconciler/metricpipeline`, `internal/reconciler/tracepipeline` (status)**: Changes `setFlowHealthCondition` errors from being joined into `allErrors` (which blocked `Status().Update`) to being logged directly, so a flow health probing failure no longer prevents the pipeline status from being written.
- **`internal/reconciler/logpipeline/fluentbit`, `internal/reconciler/logpipeline/otel`, `internal/reconciler/metricpipeline`, `internal/reconciler/tracepipeline` (reconciler)**: Returns `ctrl.Result{RequeueAfter: 30s}` when `flowHealthProbingFailed` is true, and clamps the TLS certificate requeue-after duration to a minimum of one second to prevent zero or negative durations for already-expired certificates.
- **`internal/reconciler/telemetry`**: Fixes the Telemetry reconciler to return early on error before evaluating the requeue condition, and replaces `Requeue: bool` with `RequeueAfter: 30s` when the state is `Warning`.
- **`Makefile`**: Adds a `docker-build-local` target that builds the manager image and imports it into the local k3d cluster.

</details>

## Notes for Reviewers

The self-monitor readiness gating distinguishes two cases: `ErrDeploymentNotFound` (self-monitor not deployed — no error, no requeue, `FlowHealthy` stays `Unknown`) and any other error such as `ErrDeploymentFetching`, `PodIsPendingError`, or `RolloutInProgressError` (self-monitor exists but is not ready — error logged, reconciler requeues after 30 seconds). Verify that the `errors.Is` unwrapping correctly identifies `ErrDeploymentNotFound` through the `fmt.Errorf("...: %w", err)` wrapping in `CheckSelfMonitorReadiness`.

The flow health condition error-handling change is a behavioral shift: previously, a `setFlowHealthCondition` error joined `allErrors` and caused `Status().Update` to be skipped entirely. Now the error is only logged and the status update always proceeds. Confirm this is the intended behavior and that silencing the flow health error does not mask anything more serious.

The `max(time.Until(...), time.Second)` guard is defensive: `time.Until` returns a negative or zero duration if the certificate expiry is already in the past, which would cause the reconciler to requeue with an invalid duration.

## Release Notes Input

None.

- [ ] 🔄 Regenerate and Update Summary

<details>
<summary>PR Bot Information</summary>

**Version:** `1.22.4`

- Correlation ID: `dab31650-0529-424d-93fb-1f511232c254`
- LLM: `anthropic--claude-4.6-sonnet`
- Output Template: [PR Template File](https://github.com/kyma-project/telemetry-manager/blob/0f89b291ca74aa88871f5ff518e37e3ea6528401/.hyperspace/pull_request_bot_summarize_output_template.md)
- Summary Prompt: [PR Prompt File](https://github.com/kyma-project/telemetry-manager/blob/0f89b291ca74aa88871f5ff518e37e3ea6528401/.hyperspace/pull_request_bot_summarize_prompt.md)
- File Content Strategy: Full file content
- Event Trigger: `pull_request.edited`
</details>


internal/reconciler/logpipeline/fluentbit/reconciler.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/kyma-project/telemetry-manager/internal/validators/tlscert"
 )
 
+const (
+	requeueDelayOnFlowHealthProbingFailure = 5 * time.Second
```
jeffreylimnardy: isn't 5 seconds a bit too much? why 5 seconds? and do we set this to 5 seconds on flow health probing failure or self monitor probing failure?

please extract this to a common constant unless there's a need for specific requeue durations for each reconciler

hisarbalik: We requeue when health probing failed, not self-monitoring probing.

hisarbalik: I think 5 seconds is not long, health probing can fail, when some network issues occur or timeouts

internal/reconciler/logpipeline/fluentbit/status.go
```diff
@@ -148,6 +151,23 @@ func (r *Reconciler) evaluateFlowHealthCondition(ctx context.Context, pipeline *
 		return metav1.ConditionFalse, conditions.ReasonSelfMonConfigNotGenerated, nil
 	}
 
+	// Check if self-monitor deployment is ready before attempting to probe
```
jeffreylimnardy, line 158: this code block is repeated everywhere, do we not want to extract this to something common?

internal/reconciler/telemetry/reconciler.go
```diff
@@ -135,9 +135,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 	}
 
-	requeue := telemetry.Status.State == operatorv1beta1.StateWarning
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 
-	return ctrl.Result{Requeue: requeue}, err
+	return ctrl.Result{Requeue: telemetry.Status.State == operatorv1beta1.StateWarning}, nil
```
jeffreylimnardy: this is going to cause a continuous reconciliation without any backoff right? it's anyway deprecated

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.21.0/pkg/reconcile/reconcile.go#L34-L39

jeffreylimnardy:
In general I think the scope of this PR either has to be separated (k8s test upgrades and the self monitor improvements), or you have to adjust the PR title and body. The introduced self monitor improvements are not part of k8s upgrade for test suites and will affect the final telemetry manager artifact.

internal/reconciler/logpipeline/fluentbit/status.go
hisarbalik: move to common

internal/reconciler/telemetry/reconciler.go
hisarbalik: Deprecated code refatored

test/e2e/logs/misc/reject_creation_test.go
```diff
@@ -261,7 +261,7 @@ func TestRejectLogPipelineCreation(t *testing.T) {
 					),
 				).
 				Build(),
-			errorMsg: "Invalid value: \"object\": 'tokenURL' must be a valid URL",
```
jeffreylimnardy, line 264: there test changes must be transferred to the other PR

internal/reconciler/logpipeline/otel/status.go
```diff
@@ -47,14 +48,20 @@ func (r *Reconciler) updateStatus(ctx context.Context, pipelineName string) erro
 	r.setAgentHealthyCondition(ctx, &pipeline)
 
 	if err := r.setFlowHealthCondition(ctx, &pipeline); err != nil {
-		allErrors = errors.Join(allErrors, err)
+		logf.FromContext(ctx).Error(err, "Failed to set flow health condition")
```
jeffreylimnardy, line 49: so if we're not going to return an error anymore here, then we don't need allErrors. 

but why are we not returning this error anymore? what happens when there is an issue in self monitor and deployment is never ready? do we just requeue reconciliation for all components every 5 seconds

internal/reconciler/commonstatus/checker.go
```diff
@@ -20,6 +22,25 @@ type ErrorToMessageConverter interface {
 	Convert(err error) string
 }
 
+// CheckSelfMonitorReadiness checks if the self-monitor deployment is ready before attempting to probe flow health.
+// It returns an error if the self-monitor is not ready, including when it's not deployed.
+// The caller should check if the error is workloadstatus.ErrDeploymentNotFound to decide whether to requeue.
+func CheckSelfMonitorReadiness(ctx context.Context, prober Prober, targetNamespace string, failureReason string) error {
```
jeffreylimnardy: why pass in failureReason when it's never used?

internal/reconciler/logpipeline/fluentbit/reconciler.go
```diff
@@ -35,6 +36,7 @@ type Reconciler struct {
 	agentApplierDeleter AgentApplierDeleter
 	agentProber         AgentProber
```
jeffreylimnardy, line 37: Can we rename this to a generic Prober? since now both self monitor (deployment) is using the same prober as fluent bit (daemonset agent)

PR #3638, opened 2026-06-11
chore: fix module manifests

rakesh-garimella:
## What Changed

Fixes the pre-release tag format across all release and module submission workflows, changing the separator from `rc.N` to `rcN` (for example, `1.2.3-rc1` instead of `1.2.3-rc.1`). This also simplifies the `submit-module` workflow by replacing the separate `release_tag` override parameter with a boolean `pre_release` flag.

<details>
<summary>Key Changes</summary>

- **`.github/workflows/build-manager-image.yml`**: Updates the tag pattern filter from `[0-9]+.[0-9]+.[0-9]+-rc.[0-9]+` to `[0-9]+.[0-9]+.[0-9]+-rc[0-9]+` so the image build triggers on the corrected tag format.
- **`.github/workflows/create-release.yml`**: Updates all pre-release tag computation loops and string assignments to use the `rcN` format; fixes the `submit-module` workflow trigger to pass `version` (including the RC tag) and a boolean `pre_release` flag instead of a separate `release_tag` parameter.
- **`.github/workflows/submit-module.yml`**: Replaces the `release_tag` string input with a boolean `pre_release` input across both `workflow_dispatch` and `workflow_call` triggers; updates version validation, Docker image checks, branch naming, and PR creation to derive the repository tag directly from `version`; adds an explicit pre-release channel guard that rejects non-`dev` channels for pre-releases.
- **`hack/module-release.sh`**: Removes the `repository_tag_override` parameter from `update-config`, `update-releases`, and `create-pr` commands; adds an `is_prerelease()` helper that matches the `rcN` format; updates `setup_folder` and `update_config` to compute the correct folder path for pre-releases; updates all format validation regexes and help text to use `rcN`.
- **`hack/release.sh`**: Updates a comment to reflect the corrected `rcN` pre-release suffix format.

</details>

## Notes for Reviewers

The core behavioral change is that pre-release tags now use the format `1.2.3-rc1` instead of `1.2.3-rc.1`. All dependent logic (image tag matching, git tag lookups, branch naming, folder creation, YAML field updates) has been updated consistently. The `submit-module` workflow previously used a `release_tag` string to override the repository tag in `module-config.yaml`; this override is now unnecessary because the full RC version string is passed directly as `version`.

## Release Notes Input

None.

- [ ] 🔄 Regenerate and Update Summary

<details>
<summary>PR Bot Information</summary>

**Version:** `1.22.8`

- Event Trigger: `issue_comment.edited`
- Correlation ID: `6045b82e-f91f-4e0b-8f63-300c75c66bad`
</details>


.github/workflows/submit-module.yml
```diff
@@ -234,26 +251,48 @@ jobs:
...
           CHANNEL: ${{ inputs.channel }}
-          RELEASE_TAG: ${{ inputs.release_tag }}
+          PRE_RELEASE: ${{ inputs.pre_release }}
         run: |
-          ../telemetry-manager/hack/module-release.sh update-releases "${VERSION}" "${CHANNEL}" "${RELEASE_TAG}"
+          if [ "${PRE_RELEASE}" = "true" ]; then
+            # For pre-releases, VERSION is the full RC tag (e.g., 1.2.3-rc1)
+            # Extract base version and pass VERSION as release_tag
+            BASE_VERSION=$(echo "${VERSION}" | cut -d'-' -f1)
+            ../telemetry-manager/hack/module-release.sh update-releases "${BASE_VERSION}" "${CHANNEL}" "${VERSION}"
```
shorim: why for the pre-release do we need to pass both BASE_VERSION and VERSION? Why can't we just pass the VERSION only? and what difference does it make if it is a pre-release?

.github/workflows/submit-module.yml
```diff
@@ -335,14 +373,24 @@ jobs:
...
+          # For normal releases: VERSION is the base version (e.g., 1.2.3)
+          # Extract base version for pre-releases
+          if [ "${PRE_RELEASE}" = "true" ]; then
+            BASE_VERSION=$(echo "${VERSION}" | cut -d'-' -f1)
+          else
+            BASE_VERSION="${VERSION}"
+          fi
+
+          # Call create-pr with base version, channel, output file, and full version (for pre-releases)
+          ../telemetry-manager/hack/module-release.sh create-pr "${BASE_VERSION}" "${CHANNEL}" "${OUTPUT_FILE}" "${VERSION}"
```
shorim: also here, why do we need to pass both BASE_VERSION and VERSION to the create-pr function?

.github/workflows/submit-module.yml
```diff
@@ -399,14 +447,22 @@ jobs:
...
           if [ "${CHANNEL}" = "experimental" ]; then
-            VERSION_TAG="${VERSION}-experimental"
-            FOLDER_PATH="modules/telemetry/${VERSION}-experimental"
+            VERSION_TAG="${BASE_VERSION}-experimental"
+            FOLDER_PATH="modules/telemetry/${BASE_VERSION}-experimental"
           else
-            VERSION_TAG="${VERSION}"
-            FOLDER_PATH="modules/telemetry/${VERSION}"
+            VERSION_TAG="${BASE_VERSION}"
+            FOLDER_PATH="modules/telemetry/${BASE_VERSION}"
```
shorim: Does this mean that both pre-release and actual releases will have the same folder? shouldn't they have different folders?

jeffreylimnardy (dismissed):

.github/workflows/submit-module.yml
```diff
@@ -102,36 +120,46 @@ jobs:
...
           fi
 
-          echo "Tag ${TAG_TO_CHECK} exists in telemetry-manager repository"
+          echo "Tag ${VERSION} exists in telemetry-manager repository"
 
       - name: Validate Docker images exist
         env:
           VERSION: ${{ inputs.version }}
-          RELEASE_TAG: ${{ inputs.release_tag }}
+          PRE_RELEASE: ${{ inputs.pre_release }}
```
shorim: Is the "PRE_RELEASE" env used in this job? if not, please remove it

hack/module-release.sh
```diff
@@ -235,13 +243,22 @@ get_version_tag() {
...
     echo "${version}-experimental"
   else
+    # For all other channels (regular, fast, dev), use version as-is
+    # This preserves -rcN suffix for pre-releases on dev channel
     echo "${version}"
   fi
 }
 
+# Check if version is a pre-release (contains -rcN suffix)
+is_prerelease() {
```
shorim, line 257: not necessary to change, but just to point it out that instead of figuring out if the version is a pre-release from this function, we could just pass down the "pre_release" input from the github workflow

hack/module-release.sh
```diff
@@ -372,14 +396,19 @@ update_config() {
...
   # If TELEMETRY_FOLDER is not provided, read from environment or compute it
   if [ -z "${telemetry_folder}" ]; then
     if [ -n "${TELEMETRY_FOLDER:-}" ]; then
       telemetry_folder="${TELEMETRY_FOLDER}"
     else
-      if [ "${channel}" = "experimental" ]; then
+      # Compute folder path based on version and channel
+      if is_prerelease "${version}"; then
+        # Pre-release: use full version including -rcN
+        telemetry_folder="${MODULE_DIR}/${version}"
```
shorim: If it is a pre-release or a normal release, we set the telemetry_folder the same way. You can see lines 410 and 414 are identical. The exception is if the channel is experimental, then we add the "-experimental" suffix.

Therefore, this if branch is not necessary and can be removed

hack/module-release.sh
```diff
@@ -372,14 +396,19 @@ update_config() {
   local version="$1"
   local channel="$2"
   local telemetry_folder="${3:-}"
-  local repository_tag_override="${4:-}"
+
+  validate_channel "${channel}"
 
   # If TELEMETRY_FOLDER is not provided, read from environment or compute it
   if [ -z "${telemetry_folder}" ]; then
```
shorim: Is there any place in which we are calling the "update-config" command with the telemetry_folder parameter already provided? If not , then we can remove this part and remove the intialization of the variable "local telemetry_folder="${3:-}""

hack/module-release.sh
```diff
@@ -596,32 +622,27 @@ create_pr() {
   echo "No existing PR found. Creating new PR..."
 
   # Determine version tag and folder based on channel
-  local VERSION_TAG
+  local VERSION_TAG=$(get_version_tag "${VERSION}" "${CHANNEL}")
   local FOLDER_PATH
-  if [ "${CHANNEL}" = "experimental" ]; then
-    VERSION_TAG="${VERSION}-experimental"
+  if is_prerelease "${VERSION}"; then
+    FOLDER_PATH="modules/telemetry/${VERSION}"
```
shorim: same here, the FOLDER_PATH is set the same way whether it is a pre-release or a normal release. Lines 628 and 632 are identical. Please remove this if branch

shorim (approved):

PR #3686, opened 2026-06-25
docs: fix trace_id key name in OTel log agent example

skhalash:
## Description

Fixes the "Original Log Message" example in the OTel log agent transformation documentation to use `trace_id` (snake_case) instead of `traceID` (camelCase).

The OTel log agent's trace parser only recognizes `trace_id` as the attribute key (see `filelog_receiver.go`). Using `traceID` in the example was misleading because the trace parser would not actually parse it.

## Changes

- `docs/user/filter-and-process/transformation-to-otlp-logs.md`: Changed `"traceID"` → `"trace_id"` in the original log message example to match the key the agent actually parses.


IwonaLanger (dismissed):

docs/user/filter-and-process/transformation-to-otlp-logs.md
```diff
@@ -11,7 +11,7 @@ The following example shows a container `myContainer` in Pod `myPod`, running in
   "level": "warn",
   "message": "This is the original message",
   "tenant": "myTenant",
-  "traceID": "123"
+  "trace_id": "123"
 }
 ```
```
a-thaler, line 16: In the resulting example at the end of the document, shouldn't we write traceId instead?

skhalash: Agree, also changed the type to string since bare integer does not make sense

a-thaler (approved):

IwonaLanger (approved):

NHingerl:
must add this to help portal before 1.67 regular release

PR #3731, opened 2026-07-22
docs: align status conditions and enums with source code

NHingerl:
## Description

Aligns documented status conditions, reason enums, and kubectl output examples with the actual source code. Each fix was verified against the corresponding reconciler and conditions package.

- `resources/04-tracepipeline.md`: Fix reason enum in sample (`ConfigurationGenerated` → `GatewayConfigured`) and add missing `TelemetryFlowHealthy` condition block
- `resources/02-logpipeline.md`: Add `GatewayHealthy` condition type to status table with note that it only appears when OTLP input is enabled
- `resources/04-tracepipeline.md`, `resources/02-logpipeline.md`, `resources/05-metricpipeline.md`: Document supported compression values (`none`, `gzip`, `snappy`, `zstd`)
- `monitor-pipeline-health.md`: Add note that LogPipeline and MetricPipeline display an additional `AGENT HEALTHY` column

Verification:

**`GatewayConfigured` reason**: The TracePipeline reconciler (`internal/reconciler/tracepipeline/status.go`) calls `commonstatus.GetGatewayHealthyCondition()` which sets `conditions.ReasonGatewayConfigured` for the `ConfigurationGenerated` condition type. The sample previously used `ConfigurationGenerated` as the reason value, which is the condition type, not the reason.

**`GatewayHealthy` for LogPipeline**: `internal/reconciler/logpipeline/otel/status.go:60-66` sets `GatewayHealthy` only when `pipeline.Spec.Input.OTLP != nil`. The condition is removed for runtime-input pipelines. Messages match TracePipeline/MetricPipeline (shared via `commonstatus/checker.go`).

**`AGENT HEALTHY` column**: `logpipeline_types.go:39` and `metricpipeline_types.go:39` define a `+kubebuilder:printcolumn` for `AgentHealthy` that TracePipeline does not have.

**Compression values**: The `Compression` field validation in the CRD (`+kubebuilder:validation:Enum`) accepts `none`, `gzip`, `snappy`, `zstd`.

- https://github.com/kyma-project/telemetry-manager/issues/3734

## Traceability


docs/user/resources/04-tracepipeline.md
```diff
@@ -40,9 +40,15 @@ status:
   - lastTransitionTime: "2024-02-29T01:18:27Z"
     message: ""
```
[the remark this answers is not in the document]
NHingerl: The empty message: "" is a pre-existing inaccuracy in the sample — it was already there before this PR. The code sets "TracePipeline specification is successfully applied to the configuration of OTLP Gateway" (and the MetricPipeline variant). I only changed the reason field here.

docs/user/resources/05-metricpipeline.md
```diff
@@ -53,9 +53,15 @@ status:
   - lastTransitionTime: "2023-12-28T11:27:04Z"
     message: ""
```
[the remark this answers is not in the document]
NHingerl: The empty message: "" is a pre-existing inaccuracy in the sample — it was already there before this PR. The code sets "TracePipeline specification is successfully applied to the configuration of OTLP Gateway" (and the MetricPipeline variant). I only changed the reason field here.

docs/user/resources/05-metricpipeline.md
```diff
@@ -53,9 +53,15 @@ status:
...
     observedGeneration: 1
-    reason: ConfigurationGenerated
+    reason: GatewayConfigured
```
jeffreylimnardy, line 56: arguably we can actually leave out the status subresource completely here, since creating a custom resource doesn't require you to set a status at all. the status is set by the controller and not the user

NHingerl: I feel this forces us to think about the purpose of this yaml example. 
The resources docs serve as API reference (reflecting the full CRD including controller-managed fields), separate from the user-facing instructions in `collecting-{logs/traces/metrics}/README.md`, which focus on what users apply.

The examples in all resources files (all pipelines, and telemetry) have the status block. If we remove it here, we should remove it in all of them. But I assume we originally had reasons to include them?

jeffreylimnardy: probably yeah, idk what the original reasons are, but i guess we can just include them too

jeffreylimnardy (approved):

nataliasitko (approved):

nataliasitko (approved):

